PR #65, opened 2022-09-30
Added German translation

soph-os:
Translated all strings and included changes of new language in other language files.


vassbo:
Thank you!

PR #123, opened 2023-04-02
update sk.json, fix dev

filiphanes:


vassbo:
I don't think chokidar does anything right now?!!

PR #222, opened 2023-07-25
🚩 Updated Polish | EN typo | DMG window section added

lubomsky:
No manual resize
Zrzut ekranu 2023-07-26 o 00 06 52


vassbo:
Great!

You don't need the window size!! The default size is 540 by 380, and it also auto scales to the background image if that's different in size.
But apparently it includes the top window bar height, so I just updated the background .png (again) in my last push, so that's why it works. I am pretty sure it looks the same if you remove the window size parameters!

lubomsky:
> Great!
> 
> You don't need the window size!! The default size is 540 by 380, and it also auto scales to the background image if that's different in size. But apparently it includes the top window bar height, so I just updated the background .png (again) in my last push, so that's why it works. I am pretty sure it looks the same if you remove the window size parameters!

Oh, that's right, I've verified that.
I thought that the new dmg png is on the main branch and it didn't help (I must stop working late)...
Window section is redundant now :full_moon_with_face:

PR #242, opened 2023-08-22
🚩 Update Polish

lubomsky:
For some reason I can no longer run app on my local env, so I'm not sure about context. 
After next release I will check it and maybe correct it ;)


vassbo:
Thanks. If you are on Linux, you have to change the startup time from 32 seconds to 130 seconds, at least that worked for me.

lubomsky:
I am using macos, and I have issues right after "npm install".
You can see what is happening after "npm install", and then "npm run start":

SCR-20230823-stzq

vassbo:
Okay, just remove the **grandiose** package temporarily on your mac, I guess it only works on some MacOS versions.

lubomsky:
Oh, I didn’t see that message.The problem is - I’ve cloned repo again on my computer, and now I cannot do npm install command successfully. Node modules package is erased after error/exception that I mentioned.But no worries - I’ll try on some windows laptop 😊 Wiadomość napisana przez Kristoffer Vassbø ***@***.***> w dniu 24.08.2023, o godz. 20:55:﻿
Okay, just remove the grandiose package temporarily on your mac, I guess it only works on some MacOS versions.

vassbo:
Okay. Please note that you will have to change the base repo from "**https://github.com/vassbo/freeshow.git**" to "**https://github.com/ChurchApps/FreeShow.git**", if you have an existing project. This can be done in the _.git/config_ file. FreeShow has moved to ChurchApps, so make sure you have cloned from the correct repo. 😊

lubomsky:
All clear - thanks :)

PR #281, opened 2023-11-12
🚩 Updated Polish

lubomsky:


vassbo:
You will need to use this going forward: https://app.transifex.com/nettbiter/freeshow/ (I saw you joined)
Because I am only updating the "en.json" file now with the new strings.

PR #332, opened 2023-12-10
Fix velocity override when only it is no index action, and check type match

ip4368:
midi value from the msg (midi in)'s velocity should not be overridden. UI is setting velocity to 0 when it is an index action.

To get midi value-match check working, writing it the other way, from msg (actual midi in) to the midi requirement.

Type wasn't checked too, and should too be checked.


vassbo:
Awesome! Thank you.

PR #384, opened 2024-01-23
Fix ProPresenter file import issue with negative unicode characters

0x7A79:
**I've made a fix to address the ProPresenter file import problem. This change fix the decoding function to properly handle negative unicode characters.**

Here's an example with some Chinese characters encoded in negative unicode.

_Before the fix:_
image

_After the fix:_
image


vassbo:
Nice. Thanks!

PR #442, opened 2024-03-18
Fix-Show-Groups-Sort

Dexiion:
Hi again!

I saw that the show groups where not sorted correctly, because they were sorted by strings and not by the number of the group name so I fixed it

This is the before and after of the change:

Before:
image

After:
image


vassbo:
Great! Thanks. :)

PR #449, opened 2024-03-23
Template not beeing applied to the first slide

Dexiion:
Hi!

Templates were not beeing applied to the first slide as you can see here

BUG_TEMPLATE_1

I removed that line and It worked (I don't really know what his line is doing 😅)

Thanks!


vassbo:
That's for applying templates set to individual slides, that has to stay. But maybe that's some of the issue.

PR #451, opened 2024-03-24
Create text boxes for every text box on powerpoint

Dexiion:
Hi guys!

On our church we have a template for power point slides likes this:

image

To replicate it on FreeShow using the PowerPoint utility we need to have as much text boxes as power point have. For example "Esto es un titulo" is a text box and "I. Amor" Is another text box, with this we can create a template on free show to move every text to the correct position as it is on PowerPoint

image

Sorry If i didn`t explain it well.

Hope there is no issue on adding this. Thanks! 😄


vassbo:
Great! Looks good to me. :)

PR #472, opened 2024-04-16
Ukrainian localization improvements

MiragePresent:
## Description

Updated Ukrainian localization. Didn't have chance to check how every line fits in the interface but still should be much better. For example "Stage" was translated as stage in the process not "Stage" for people performances.
I'm planning to live some time with this translations and fix them further if needed. I believe some of them should be translated according to the context and not literally from English to Ukrainian


vassbo:
Great thanks! Please head over to Transifex to translate going forward, as the project file does not contain all of the newest strings (I have updated it with your changes):
https://app.transifex.com/nettbiter/freeshow/language/uk/

PR #516, opened 2024-05-15
Fix: Collapsable Edit Menu Sections

djsamseng:
Test Plan: Open the Welcome slide show. Click Edit. Open and close the right hand sections such as "Align", "Text", "Lines", "Chords", "Special", "CSS". Ensure that opening the first time set the value to the default (for align this changes the text align from center to left alignment - this was the previous behavior). Ensure that collapsing the menu keeps the value it was overwritten to. Ensure that reopening the menu keeps the value to was it was overwritten to. Ensure that clicking reset converts the value back to the original value before the menu was opened (note this may be different than the default for example with Align, this was the previous behavior).

Note that "CSS" is missing from the language dictionary and thus collapsing CSS exposes a race condition bug in that CSS will now show "[edit.CSS]" instead of "CSS" upon collapse. If you have any instructions for updating the langauge strings that would be appreciated!

collapsableMenu

### DIFF
--- a/src/frontend/components/edit/tools/EditValues.svelte
+++ b/src/frontend/components/edit/tools/EditValues.svelte
@@ -225,10 +225,6 @@
     function checkIsClosed(id: string) {
         if (noClosing) return false
 
-        if ($storedEditMenuState[sessionId]?.[id] !== undefined) {
-            return $storedEditMenuState[sessionId]?.[id]
-        }
-
         let closedVal = closed[id]
         let defaultEdit = defaultEdits?.[id]
         let currentEdit = clone(edits[id])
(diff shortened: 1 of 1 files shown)


vassbo:
Thanks. But I don't see any reason why you would need to collapse them!? All of the "changed" menus are automatically opened anyway to easily see what has changed from the default.

The CSS is not a part of the language, and it has changed for some reason, `section[0]` is not upper case, but it should be. So you can test how `section` changes

vassbo:
Closing for now, as I guess we would need a new way of handling the input sections that is less confusing than the current system. 😊

PR #591, opened 2024-06-17
Add more checks on whether the note is correct

ip4368:
Not sure why on MacOS, it seems to repeat the same note for all the Midi id, and if we are not checking the channel/note, it will go crazy and execute basically available action.

This add the checks for channel and note, so that it will only fire if the channel and note are matching the action. Not sure if this is the right fix, or if something can be fix on the midi initial dispatching.

### DIFF
--- a/src/frontend/components/actions/midi.ts
+++ b/src/frontend/components/actions/midi.ts
@@ -97,7 +97,10 @@ export function receivedMidi(msg) {
 
     // the select slide index from velocity can't select slide 0 as a NoteOn with velocity 0 is detected as NoteOff
     // velocity of 0 currently bypasses the note on/off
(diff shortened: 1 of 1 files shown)


ip4368:
Probably a bit more context, the source firing for all midi id was probably always present, and it was caught by this line before the migration, which is now removed:

https://github.com/ChurchApps/FreeShow/blob/ce0141d189d8830eb03bfed2a169a091abcfd591/src/frontend/utils/midi.ts#L124

vassbo:
Great! Thanks. :)

ip4368:
My bad that I didn't follow the naming convention. @vassbo, I can clean it up or if you prefer to clean it up yourselves, please do so.

vassbo:
No worries @ip4368, shouldn't be a problem.

PR #605, opened 2024-06-22
Making test to be independent from what was on the existing setup

ip4368:
When different people trying to run the test on their own computer, it actually might or might not run depending on many factors, language selected or whether there is specific group name, etc.

Another issue is that every time a test is run, there will be an associated new presentation that need to be manually deleted later on, as we also didn't do delete the created presentation on the UI.

To solve both of the problem, I am trying to use tmp directory for both setting and presentation data, so that it will always start clean.

This should make the test more stable across computers.


vassbo:
Perfect! Thank you. 😃👍

PR #606, opened 2024-06-22
Add playwright config and github worklow to run e2e test

ip4368:
After landing the machine independent e2e test, it enables us to start doing CI on github, as it won't matter with previous setup.

Having the ability to mock out the settings and show data storage, also allows us to test with different preset settings and show data, this allows us to test a few things
- Old settings/show data won't be broken with newer code/testing if migration works.
- Making sure no unexpected UI is showing up with specific settings, i.e. start up screen should show up with old settings.

This allows us to test many more things and will ensure the quality of releases.

### DIFF
--- a/.github/workflows/playwright.yml
+++ b/.github/workflows/playwright.yml
@@ -1,0 +1,34 @@
+name: Playwright Tests
+on:
+    push:
+        branches: [main, dev]
+    pull_request:
+        branches: [main, dev]
+    workflow_dispatch:
+
+jobs:
+    test:
+        timeout-minutes: 60
+        runs-on: ubuntu-latest
+        steps:
+            - uses: actions/checkout@v3
+            - uses: actions/setup-node@master
+              with:
+                  node-version: 18
+            - name: Install dependencies
+              run: npm install
+            - name: Install Playwright Browsers
+              run: npx playwright install --with-deps
+            - name: Build
+              run: npm run build
+            - name: Run Playwright tests
+              run: xvfb-run --auto-servernum --server-args="-screen 0 1280x960x24" -- npm run test:playwright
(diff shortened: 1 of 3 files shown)


vassbo:
@ip4368 Do you have any idea why the Playwright test suddenly times out on teardown? It just started to happen seemingly when not much was changed. All the tests are passing! After some configurations I tried running on Windows, and on Linux VM, and that stops the tests successfully, but the GitHub Actions fails!
Error page: https://github.com/ChurchApps/FreeShow/actions/runs/9987678534/artifacts/1714067472

ip4368:
@vassbo let me debug a bit today

vassbo:
Thank you! Now it suddenly passes again, I think something weird happend on the Actions backend, and I have no idea why!

vassbo:
And now it stopped working again, please let me know if you figure something out.

ip4368:
I think the electronApp.close() need to be awaited, but I don't think that would change drastically, as sometimes it still fires the awaitables. What was the issue of awaiting that and needed to be removed? The .close() is causing timeout, then something is wrong there too.

vassbo:
Okay, the electronApp.close() was awaited in the first place, but never exited suddenly, so I tried removing it and that seemed to help a bit. Still not sure though.

ip4368:
I tried adding the await back and the test on github action seems to be timing out awaiting the app to be closed even when I commented out every single line of test. My linux desktop is in storage but I am planning to set it up again. Let me see if I can reproduce it. I am guessing something is causing the app to not be fully shutdown cleanly.

ip4368:
Was a bit tricky to bisect if as the dev branch was refreshed and losing the record on when it started to happen.

PR #635, opened 2024-06-30
implemented rudimentary support for WebRequests

thegreeny:
implemented rudimentary support for WebRequests.
Tested against https://httpbin.org
Also tested and running locally.

Use-Case:
In our church we use a phillips-hue-bridge, which is controllable over simple WebRequests.
Now its possible to automaticaly dimm the main lights, when a Video is played.


vassbo:
Awesome! 😃

PR #669, opened 2024-07-18
Add Songbeamer import

ItsFlo:
This also adds changes to some existing code:
- Update the `T` svelte component when the id changes
- Changed the category sorting in the CreateShow Popup to use the translated names of the default categories
- Add chords with a negative pos at the beginning of lines in the stageview (Songbeamer can have pos -1)
- `startImport` now forwards all data to `importShow` (I needed this for the Songbeamer importer)

I extracted the icon from the SongBeamer.exe file.

This adds a new option on the import popup that opens a new popup with a few import settings.\
The encoding is necessary because Songbeamer changed the encoding of their files in version 6.02a.\
Songbeamer can have multiple languages so I added three different options how to import them.


vassbo:
Great! Thank you. :)

vassbo:
You will probably want to translate the German strings here so they don't get overritten: https://explore.transifex.com/nettbiter/freeshow/

Issue #699, opened 2024-07-28
Motions background freezes after upgrade to new version


odlarhg:
Hello

After updating to the new version, the motions videos freeze momentarily during their execution. The video does not look continuous, it has small freezes during its execution. I have just updated yesterday to the new version and before updating the video was running normally.

I am using the same computer. My computer has Core i5 8th gen, 16gb RAM.

Thanks for your help.

jzongker:
Could you let me know if you're running a single output or multiple screens?

odlarhg:
Hello

I am using an Epson powerlite x36+ projector to project the lyrics and my laptop screen.

odlarhg:
Hi

Attached is a sample of the video playing as a background in FreeShow and the same video playing with the Windows 11 media player.

As you can see, the execution speed is very different.

Before updating I had version 1.1.9 installed, in that version the video played correctly in FreeShow, but after the update it no longer plays correctly.

I am running these video samples using only my computer screen.

https://github.com/user-attachments/assets/98ed5376-4b5d-49ef-a496-9f3585fc1ae1
https://github.com/user-attachments/assets/16b51197-f36f-43f3-90d4-971860b30cfc

vassbo:
Thanks @odlarhg. Can you please try this preview build and let us know if it's better: https://filetransfer.io/data-package/T5OlByGl

vassbo:
I noticed that it's fine until the video loops. Then it starts to stutter until the video is paused and played again!

PR #701, opened 2024-07-29
Made preview video time syncing less aggressive

jzongker:
The preview video time stamp is pretty much always a few milliseconds off from the main output.  By trying to bring it back in perfect sync we're making the problem worse because it stalls for a moment when skipping ahead or back, putting it further out of sync.  Doing this dozens of times per second causes the video to stutter.  I updated it to only do this if we're at least 500ms out of sync.


jzongker:
I suspect this is also the cause of issue #699 but haven't been able to reproduce the issue to test.

vassbo (approved):

PR #712, opened 2024-08-01
Added python installation instructions to building instructions

thebiblelover7:
Quick pull request to add python installation instructions. I found that when I had python3.12 installed, I couldn't build FreeShow until I had the `setuputils` python package. Hopefully it helps someone later on :).


vassbo:
Nice. Thanks!

PR #723, opened 2024-08-07
NDI Inputs

vassbo:


Marsipulami:
Hi.

this update implies that it is possible to show NDI sources elsewhere on the network right?
How to enable or configure these sources?

Is there already some sort of documentation?

Thanks
Nathan

vassbo:
Hi, not much documentation about this right now. Try going to "Media">"Screens">"NDI" you should find the inputs here.

Issue #353, opened 2023-12-30
TRANSITIONS BUG


SKU003:
Hi, I don't know what can cause it, but it's like the transitions works when it wants too. Cause when I wanted to take footage of it, it started working again even when I closed and opened it again it worked but honestly most of the time it never works even when I try other transitions and this is both for text and Media transition. Because I'm sure the next time I open it, it won't work.

vassbo:
I'll see if I can figure it out. Hiding and showing the outputs again after the app has started should make the transitions work always.

vassbo:
Let me know if you still experience this in the latest version. I'm not sure what causes it.

lwallace22:
could you advise which version does not have this issue ?

vassbo:
@lwallace22 Looks like it happens sometimes for some users in all versions.

PR #854, opened 2024-09-27
v1.2.8-beta.2

vassbo:


gkiibzmultimedia:
the transitions is not working in this version

vassbo:
@gkiibzmultimedia In my testing it works fine, make sure you have not set any custom item, or slide, or style transitions, then change the global transition to e.g. "Fade" and 0.5 seconds, and make sure that works!
bilde

gkiibzmultimedia:
I revert back to beta 1 and the transitions are working fine.

vassbo:
Looking more into it, apparently it's not unique to the latest pre-release as it have probably been an issue for a long time.
The same issue I think: https://github.com/ChurchApps/FreeShow/issues/353

It must be a Svelte bug, and it seems to happen like 50% of the time, so restarting the app (output window) will eventually make it work.
The strange part is that it works 100% in the development build. And like 80% of the time in the preview.

Apparently if the output window is displayed right after opening, it works more of the time I think.

lwallace22:
video provided i see no solution to this issue

PR #1016, opened 2024-11-13
chore: non-conflicting dependency updates

DAcodedBEAT:


vassbo:
I don't think we need package-lock in the source, also have you tested e.g. the Google Drive sync?

DAcodedBEAT:
@vassbo `package-lock.json` is important, especially in projects with contributors, since it ensures everyone is fetching the same dependencies.  Here is a great explanation as to why package-lock should exist: https://blog.logrocket.com/why-you-should-use-package-lock-json/

As for testing all of the flows, I figured this would be fine since typescript passes, meaning none of the function signatures have changed.

vassbo:
I see.. Probably fine!

@jzongker What do you think? Is there a reason you removed the package-lock?
https://github.com/ChurchApps/FreeShow/pull/291/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947R9

jzongker:
Adding it back is probably the right call.  I've typically excluded it to allow the packages to be updated with bug fixes and non-breaking changes, but the risks of the production builds being slightly different code than development likely outweighs the benefits.

vassbo:
Sqlite3 must stay at 5.1.6:
- https://github.com/ChurchApps/FreeShow/commit/c45907a8974bdd79268b985d3dcb1160d17502e2
- https://github.com/TryGhost/node-sqlite3/issues/1746#issuecomment-1895668411

But the rest looks to be working as it should!

DAcodedBEAT:
> Sqlite3 must stay at 5.1.6:
> 
> * [c45907a](https://github.com/ChurchApps/FreeShow/commit/c45907a8974bdd79268b985d3dcb1160d17502e2)
> * [I couldn't build successfully in electron TryGhost/node-sqlite3#1746 (comment)](https://github.com/TryGhost/node-sqlite3/issues/1746#issuecomment-1895668411)
> 
> But the rest looks to be working as it should!

@vassbo Did you experience that locally? I was able to run the playwright test, which made me think this should work.

PR #1031, opened 2024-11-20
Update snapcraft.yaml

eeickmeyer:
This actually gives the snap access to /usr/share/alsa/alsa.conf, but doesn't completely resolve the MIDI access causing it to crash. It also beautifies snapcraft.yaml. One step at a time.


eeickmeyer:
I also just moved it to core24 because I figured we might be able to get better alsa support that way.

PR #1087, opened 2024-12-06
chore(codestyle): fixup codebase to more-modern practices

DAcodedBEAT:
Most prominent changes:

1. **Prefer `const` over `let`**  
   - Signals intent: the variable won’t be reassigned.  
   - Improves code clarity and reduces bugs.  

2. **Use `node:*` for Node.js imports**  
   - Clearly indicates built-in modules.  
   - Avoids naming conflicts and aligns with modern standards.  

3. **Use `import type` for importing types**  
   - Ensures no runtime code is imported.  
   - Reduces bundle size and improves clarity.  

4. **Prefer arrow functions**  
   - Cleaner syntax and auto-binds `this`.  
   - Reduces boilerplate and common errors.  

5. **Break up long lines**  
   - Easier to read and understand.  
   - Simplifies maintenance and debugging.  

### Summary  
These changes enhance readability, clarity, and alignment with modern best practices, making the codebase cleaner and less error-prone.


vassbo:
- Test failed as install does not work.
- Also prettier should be updated if semicolons are a standard
- I don't see the reason for node: ?
- Also most IDE's split large lines into multiple more readable lines

vassbo:
Actually looks like it failed because NDI 6.1 was just released yesterday.

DAcodedBEAT:
@vassbo
* correct, the tests failing seem to be related to NDI and not the javascript changes itself
* I can remove the semicolons if you prefer that
* the `node:*` prefix for builtin modules helps optimize the import and prevent potential typosquatting attacks (accidentally installing a name which collides with a builtin and using the wrong module)
* sure, but not everyone uses an IDE - I'm thinking of the code review process but there might be others which use standard text editors.

vassbo:
Okay, I think we should keep the semicolons disabled for now as that is how the code has been for now, and I'll try to merge it afterwards, but it seems like this prettier config is ignored as semi is disabled here:

```json
    "prettier": {
        "trailingComma": "es5",
        "printWidth": 250,
        "tabWidth": 4,
        "semi": false,
        "singleQuote": false
    }
```

PR #1155, opened 2024-12-30
Clock Format

chalamministries:
Added ability to add a DATE to the Clock Item.
Added ability to choose from a range of different formats.
Added ability to set a custom format based off DayJS.

Other improvements:
Updated itemHelpers.ts to utilize default styles set in the boxes.ts configuration so Item's display with the "default" styles.


vassbo:
I apparently merged right to the dev branch, but I got the conflicts resolved there.

I found a bunch of things that should be changed:
- The 24 hour format set in the settings should be used for the preset clock values, so if 24 hour is enabled that should be the default clock. We do not need the other options in the dropdown, 12/24 time format should only be set in the custom input, as this is almost never used when set already in the settings.
image
- I'm not sure if monospace is the right default font to use for the clock.
- The analog clock does not work!
- The drawer clock should be reverted to how it was. (With the style & settings time format set)
image
image
- Stage clock style does not work!
- Custom format tip should be done in a better way, example: Hover over "Default project name" in the settings to get similar values.

PR #1182, opened 2025-01-11
Beblia Import - Single Verse Fix

chalamministries:
Chapters with one verse were not reading as an array


vassbo:
Too many commits, I will just update the file manually on the dev branch as it's just one file changed. I don't think you reset the dev branch before creating this commit.

PR #1195, opened 2025-01-15
Snap app packaging tweak: Add removable-media to snapcraft.yaml

navhaxs:
Users who install FreeShow via Snap package (i.e. Ubuntu Store) cannot import/export files to removable media due to the sandboxing.

Fix:
This PR adds "removable-media" to `snapcraft.yaml` so that we (as the publisher of the FreeShow package) indicate that the application _can_ use removable media.

Note:
The user may still need to manually enable the permission, but at least they will now have "removable media" as a permission they can enable. See https://askubuntu.com/a/1267296 and https://snapcraft.io/docs/interface-management#heading--auto-connections

### DIFF
--- a/snap/snapcraft.yaml
+++ b/snap/snapcraft.yaml
@@ -54,6 +54,7 @@ apps:
       - audio-playback
       - pulseaudio
       - opengl
+      - removable-media
     environment:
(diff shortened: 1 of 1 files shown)


vassbo:
Awesome! Thanks.

PR #1261, opened 2025-02-09
feat: Adds an option to toggle date on Stage

jeffreypaul15:
Hey team!
We recently started using FreeShow as an alternative and found that this option was quite necessary for us.
Please consider this change and I also needed a bit of help regarding the translations for this feature.
Thanks!


chalamministries:
I did the date part for the main slides. Kristoffer was going to do the stage display. But can you create the pull request on the dev branch instead of Main?

vassbo:
We can do this temporarily until I get the stage upgraded.

PR #1486, opened 2025-04-15
retry-after should be lowercase

zefanjajobse:
I tested the dev branch for the rate-limit issue of planning center, it does not use the retry-after from the header. Apparently the header is lower-case.

image


vassbo:
Cool, thanks! Glad you spotted that.

PR #1502, opened 2025-04-20
Update npm packages

Auxority:


Auxority:
Do not merge please - it's in a broken state right now. But I don't see any errors anymore, only ARIA warnings from usability issues.

Auxority:
Todo: rebase to dev branch

Auxority:
I think this PR has too many issues, will try to do it in smaller steps.

Auxority:
@vassbo I only added the “any” annotations in places where there weren’t types defined before. It would make this upgrade a 4-week process instead of a 3-hour process.

I just added @ts-ignore to places where it worked before but where the fix for the new TS version required some more work.

It is very easy to create issues from these through a simple find query later, but not worth it to do that all at once.

Auxority:
Will check the right click and transitions issue.

vassbo:
Okay, is the new typescript version requiring any for all unset values?

Auxority:
Yes, the project wouldn’t compile without it. It would just show those as an error.

vassbo:
Maybe it's possible to change a value in the tsconfig to fix that.

Auxority:
That’s a good idea actually.

Auxority:
Just checked the repository, but there are 1878 `any` types declared already on the `dev` branch.

I know this is far from ideal, but adding a couple more due to the TypeScript upgrade isn't going to hurt. By disabling the `noImplicitAny` check in the tsconfig we wouldn't know about the issue at all. This at least allows us to look for `: any` in our code and refactor where needed.

vassbo:
I get that there are a lot of `any` types, way more than it should be (I actually replaced a couple hundred of them with proper types a few weeks ago).
But the problem is that many of the values you set to `: any` was typed indirectly already based on a parent type or on itself, so we don't want to "remove" the existing type. (There is a setting that can be enabled to see them)

Here are a few examples (there are many more):
image
image
image
image

Auxority:
It’s strange that VSCode doesn’t recognise that for me. But good to know :)

PR #1530, opened 2025-04-25
Fix makesnap for commit that organized the scripts

eeickmeyer:
The initiative to organize the scripts broke the snap building process. Snaps require a very specific structure and the `snap` directory _must_ be in the root of the repository. WIthout this, it fundamentally breaks.

Additionally, the removal of the executable bit from the scripts caused the entire build to fail as the scripts _must_ be executable in order for the build to succeed and the installed snap to function at all.

I have revised the `makesnap` script to accommodate the changes and move the required files to a `snap` directory during build so it may build properly.

### DIFF
--- a/makesnap
+++ b/makesnap
@@ -5,6 +5,11 @@ version=$(grep "\"version\"" package.json | cut -d ":" -f 2 | cut -d "\"" -f 2)
 sed -i "s/version:.*/version: ${version}/" config/building/snapcraft.yaml
 npm install
 npm run build
-npm run pack
+npm run release
+chmod +x scripts/snap/*.sh
+mkdir snap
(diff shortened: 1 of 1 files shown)


vassbo:
Sorry, I did not realize that. I noticed the "pack" command being used afterwards, so I added that back on the dev branch if that's better than "release". Thank you for fixing it!

eeickmeyer:
> Sorry, I did not realize that. I noticed the "pack" command being used afterwards, so I added that back on the dev branch if that's better than "release". Thank you for fixing it!

Yeah, "pack" definitely works better as it doesn't create the appimage or anything else, just what becomes the final product, which is exactly what the snap needs.

PR #1557, opened 2025-04-30
Disable echo cancellation for microphone audio inputs

lehik:
The proposed changed disables it for all mics. Maybe it could be done per-device if anyone actually uses such audio inputs where echo cancellation is needed.

### DIFF
--- a/src/frontend/audio/audioMicrophone.ts
+++ b/src/frontend/audio/audioMicrophone.ts
@@ -21,7 +21,7 @@ export class AudioMicrophone {
         }
 
         navigator.mediaDevices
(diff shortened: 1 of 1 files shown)


src/frontend/audio/audioMicrophone.ts
```diff
@@ -21,7 +21,7 @@ export class AudioMicrophone {
         }
 
         navigator.mediaDevices
-            .getUserMedia({ audio: { deviceId: { exact: id } } })
+            .getUserMedia({ audio: { deviceId: { exact: id }, echoCancellation: false } })
```
vassbo, line 24: @thebiblelover7 Here's the fix we were looking for. Simpler than we thought.

PR #1580, opened 2025-05-04
slightly refactor loading of Planning Center services

Auxority:
Hey there, just thought it would be nice to apply some slight refactoring here to make it easier to maintain.

I was attempting to fix an issue, but wasn't able to figure it out. Maybe someone with more knowledge on the project can help out. For some reason items/songs with the same name are suffixed with a number, even though they are in different Planning Center Plans / FreeShow shows.

The first import goes perfectly fine:
image

Second one has one minor issue:
image

But a service with the same template gets quite messy:
image

If anyone can hint me in the right direction I am more than happy to create a PR for a fix.

### DIFF
--- a/src/electron/planningcenter/request.ts
+++ b/src/electron/planningcenter/request.ts
@@ -80,120 +80,238 @@ export async function pcoRequest(data: PCORequestData, attempt = 0): Promise<any
 // LOAD SERVICES
 
 const ONE_WEEK_MS = 604800000
+
 export async function pcoLoadServices(dataPath: string) {
+    const serviceTypes = await fetchServiceTypes()
+    if (!serviceTypes.length) return
+
+    sendToMain(ToMain.TOAST, "Getting schedules from Planning Center")
+
+    const { projects, shows, downloadableMedia } = await processServiceTypes(serviceTypes, dataPath)
+
+    downloadMedia(downloadableMedia)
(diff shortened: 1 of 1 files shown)


vassbo:
Cool, looks good.

The numbers are added because by the looks of it Planning Center are setting a different ID to the songs when they are in different services, even though they are the same.

Because of that the file names can't be the same in FreeShow becuase that does not work with the file system, so a fix would be to mark all imported Planning Center songs with a origin source as "PCO" or something, so we can on each sync check all songs with that mark and with the same name; if it's the same content, we replace the project id by the existing one and don't import the new  song that is identical.

Auxority:
Thanks, that explanation really helps. I’ll try it out today and see if that is indeed the issue :)

vassbo:
All the .show files are stored in the same folder and I don't see any good reason to change that.

If it's the same song, but a different arrangement we can create a new layout in the existing show with the same name.
But if a show already exist with the same name, but it's not originally from Planning Center, that's when we should add the " 2" because it should be a new show. If that makes sense.

Auxority:
That makes sense, I wouldn't make a change in the directory structure then.

I like the idea of using a "show" for a "song" and "layouts" for "arrangements", keys are not really relevant for lyrics anyways. I'll do some tests :)

vassbo:
I added the origin key now, so all Planning Center songs will be marked with the key `"origin": "pco"`

PR #1582, opened 2025-05-05
fix duplicate shows from Planning Center items

Auxority:


vassbo:
I don't see how this makes much of a difference, if the ID from Planning Center already is unique?!

Auxority:
The thing is, we don’t always want them to be unique. And we can’t just use the CCLI Song ID and Arrangement ID, because someone might change the lyrics and then the show wouldn’t update right?

Auxority:
Ahh alright, I didn't know that. Let me simplify it then :)

Auxority:
The item fix might not work for everyone. If people do actually use items with the same with different content for each service, then we'll have to use item IDs. But that would mean that a new show is created for every item every week, even if they share the same title and type.

Auxority:
I tested it locally and this fixes the issue I was having. We were using the SongItemID (which is always unique - not related to the SongID), which I have now changed to the SongID and ArrangementID.

Auxority:
Ideally we'd merge the different arrangements of a CCLI Song Number (or even better the PC Song ID) into the same show, but I have no idea how to do that.

Auxority:
Is it just a matter of adding a layout per arrangement here? (And replacing the layoutId with the arrangementId?)

```ts
const layoutId = uid()

    const show: Show = {
        name: title,
        category: "planning_center",
        timestamps: { created: new Date(SONG.created_at).getTime() || Date.now(), modified: new Date(SONG.updated_at).getTime() || null, used: null },
        meta: metadata,
        settings: {
            activeLayout: layoutId,
            template: null,
        },
        layouts: {
            [layoutId]: {
                name: "Default",
                notes: SONG.notes || "",
                slides: layoutSlides,
            },
        },
        slides,
        media: {},
    }
```

vassbo:
That is what we need to do yes, but we would also only want to add new arrangements, and not replace existing ones. Also we should merge groups that have the same content into the same group, so it's a bit more work.

Auxority:
Oh that's a good point. Shall I make an issue for it? I don't have time to develop it anymore today.

vassbo:
We probably do not need an issue for it, but we can keep it in mind and see what to do in a while, if the current importing becomes a problem.

Auxority:
Alrighty, will add it to my personal notes for later :)

PR #1688, opened 2025-05-30
Allow user to select which categories contain songs to sync to chums

jzongker:
This is my first time creating a new settings field.  Please check my work.

image


vassbo (commented):
You did the new settings key correctly!

The data should be sent from the frontend though as the stores.SETTINGS might not have been saved yet.
I also noticed that you used the default documentsFolder() path for the showsPath, but that might have been changed to a different location, so we should pass that from the frontend as well.

But I made these fixes for you, you can see what I changed if you want.

vassbo:
I also updated the popup UI to match the rest of the app:
image

PR #1750, opened 2025-06-17
fix(text-editor): prevent chord movement corruption in text editor

gladsonsam:
Fixed two bugs in slide matching logic that caused text corruption when moving chords:

1. Slide fingerprinting: Create chord-position-agnostic fingerprint using sorted unique chords instead of position-dependent chord order
2. Content update: Always update slide content when matched, not just when slide count changes

This ensures chord movements like \Ama[G]zing\ → \[G]Amazing\ preserve all content correctly.

 What Was Happening Before:

  1. User edits: Ama[G]zing → [G]Amazing
  2. Slide matcher: ❌ Fails to match (due to chord positions) → creates new slide → loses content
  3. OR Slide matcher: ✅ Matches → but content never updated → reverts to old content

 What Happens Now:

  1. User edits: Ama[G]zing → [G]Amazing
  2. Slide matcher: ✅ Matches (chord-position-agnostic fingerprint)
  3. Content updater: ✅ Updates slide content (always updates when matched)
  4. Result: [G]Amazing gr[G/B]ace how swe[C]et the sou[G]nd ✅

### DIFF
--- a/src/frontend/components/show/formatTextEditor.ts
+++ b/src/frontend/components/show/formatTextEditor.ts
@@ -4,7 +4,7 @@ import type { Chords, Item, Line, Show, Slide, SlideData } from "../../../types/
 import { activeShow } from "../../stores"
 import { createChord } from "../edit/scripts/chords"
 import { DEFAULT_ITEM_STYLE } from "../edit/scripts/itemHelpers"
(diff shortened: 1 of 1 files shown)


vassbo:
Cool, this seems like it's working fine.

Issue #1653, opened 2025-05-20
Add text input as import method


gladsonsam:
**Describe the feature**
Can we add a new input method or replace the clipboard with text import, where it will open a text box where I can paste text? Not sure if it's just me that would like this, as I don't use the clipboard import method. The idea behind this is that I can then paste lyrics or even lyrics with chords:

 
[Verse]
C                  D  C
All the saints and angels
     C               D  C
They bow before Your throne
C              D  C
All the elders cast their crowns
           C       D       G
Before the Lamb of God and sing
 
 
[Chorus]
                      G                         D
You are worthy of it all, You are worthy of it all
                     C                          D
For from You are all things, and to You are all things
                  G
You deserve the glory

Then a parser could convert it into FreeShow text:

[Verse]
Al[C]l the saints and an[D]gel[C]s
They bo[C]w before Your th[D]ron[C]e
Al[C]l the elders ca[D]st [C]their crowns
Before the La[C]mb of Go[D]d and si[G]ng

[Chorus]
You are worthy of it all[G], You are worthy of it all[D]
For from You are all th[C]ings, and to You are all th[D]ings
You deserve the glor[G]y

Currently, I use a simple app I made to do this for me (https://github.com/gladsonsam/SongBuilder), but I would rather do this all within FreeShow to make it easier for volunteers.

vassbo:
You can paste any text when you click "New show", and choose "Quick lyrics"!

gladsonsam:
Sorry, I forgot about that. What about the parsing ability? Can we maybe add an option under "format text" called add chords, Then it would accept this:

 
[Verse]
C                  D  C
All the saints and angels
     C               D  C
They bow before Your throne
C              D  C
All the elders cast their crowns
           C       D       G
Before the Lamb of God and sing
 
 
[Chorus]
                      G                         D
You are worthy of it all, You are worthy of it all
                     C                          D
For from You are all things, and to You are all things
                  G
You deserve the glory

Forgive me if this already exists somewhere!

PR #1754, opened 2025-06-18
Improve Chord Parser in Quick Lyrics

gladsonsam:
Sorry, still new to coding, so I'm not sure if I did the request right. The only edit is to src/frontend/converters/txt.ts

Marking as draft, as I haven't created an issue (more of a continuation of [#1653](https://github.com/ChurchApps/FreeShow/issues/1653#issuecomment-2894321195)). Currently, I can't paste songs from places like [Ultimate Guitar ](https://tabs.ultimate-guitar.com/) or [pnwchords](https://pnwchords.com/) without some sort of issue, such as improper chord placement or blank slides.  

This improved chord parsing and has worked for songs from both Ultimate Guitar and pnwchords.

I haven't tested normal songs or other inputs, as I don't know how else people use Quick Lyrics. Let me know if any changes are needed.

### DIFF
--- a/src/frontend/converters/txt.ts
+++ b/src/frontend/converters/txt.ts
@@ -49,14 +49,10 @@ export function convertText({ name = "", origin = "", category = null, text, noF
     // in "Text edit" spaces can be used to create empty "child" slides
     text = text.replaceAll("\r", "").replaceAll("\n \n", "\n\n")
 
-    // preprocess chord lines before splitting into sections (only if formatting is enabled)
-    const formatText: boolean = noFormatting ? false : get(formatNewShow)
-    let processedText = text
-    if (formatText) {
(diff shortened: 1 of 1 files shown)


gladsonsam:
https://github.com/user-attachments/assets/b472af9f-521a-497b-979d-89875ec452c2

vassbo:
Seems good. But I think we can extract the chords even when formatting is not enabled. Or what do you think?

gladsonsam:
Seems fine to me. Don't know if there is an edge case where someone doesn't want it formatted or in case the chord line detection doesn't work, but it works fine for me.

vassbo:
Okay, would you like to change that? Other than that it seems fine.

gladsonsam:
Should be all good now

Issue #1763, opened 2025-06-19
FS is freezing when we click on Stage Tab on v1.4.6-beta1


iAmCLY:
**Describe the bug**
FS is freezing when we click on Stage Tab.

**Version (Optional)**
 - OS: Windows
 - FreeShow Version: 1.4.6-beta1

vassbo:
Looks like the fix is to create (and then you can remove it) a RSS value in settings>General>Manage dynamic values.

iAmCLY:
> Looks like the fix is to create (and then you can remove it) a RSS value in settings>General>Manage dynamic values.

Yes, that works as work around

PR #1764, opened 2025-06-20
Fix TypeError when dynamicRSS is undefined in getDynamicIds

gladsonsam:
Resolves error: "Cannot read properties of undefined (reading 'length')" when get(special).dynamicRSS is undefined by ensuring rssValues is always an array using nullish coalescing operator.

Looks like It fixes this error: https://github.com/ChurchApps/FreeShow/issues/1763


vassbo:
Thank you!

PR #1774, opened 2025-06-24
Fix/rtl output screen

minaalfy:
This is a fix for RTL in output screen.


vassbo:
How did this make a difference when we have the exact same code in the "App.svelte" file?

vassbo:
I think you would want to use the `listeners.ts` and `receivers.ts` file to send those values to the output window.

PR #1775, opened 2025-06-24
add arabic bible books names

minaalfy:


src/frontend/converters/bebliaBible.ts
```diff
@@ -169,3 +172,74 @@ export const defaultBibleBookNames: any = {
     65: "Jude",
     66: "Revelation"
 }
+
+export const arabicDefaultBibleBookNames: any = {
```
minaalfy, line 176: @vassbo It would be preferable to provide this as an API rather than including it in the main bundle, especially if we want to support additional languages in the future.

vassbo:
The Bible book names should be included by the Bible provider in the XML file, because there are so many different names out there, sorry.

vassbo:
Even one language can have multiple variations of the name in different versions.

minaalfy:
The main Arabic bible version isn't in the API and available only from Biblia which doesn't have book names and that's why you created the `BibliaBible.ts` and added a default book names in English.

vassbo:
Yeah, thats' a problem with many Beblia XML files. You or the distributor needs to add the book names to the XML file! In the book tags: `name="Genesis"`! I only added the English because that's better than nothing, or just showing a number at least.

PR #1787, opened 2025-06-27
Get Variables API

chalamministries:
Added get_variables call to return all variables
Added get_variable {"id?": string, "name?": string} to search for variable

### DIFF
--- a/src/frontend/components/actions/api.ts
+++ b/src/frontend/components/actions/api.ts
@@ -37,7 +37,9 @@ import {
     getShow,
     getShowLayout,
     getShows,
-    getSlide
+    getSlide,
+    getVariables,
+    getVariable
 } from "./apiGet"
 import {
     addGroup,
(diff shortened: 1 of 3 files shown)


vassbo:
You can actually get a variable value with the get_dynamic_value action. But I think it's good to have this as well.

It looks fine, but a couple of notes:
- You should extend the type for `getVariable`/`get_variable` in `api.ts` (instead of `{ id?: string; name?: string }`) like the other values, because that's parsed and placed on the api docs website.
- We can remove the `Variables` type as it is not in use.

chalamministries:
Yes. I was going to use the Variables type in the beginning but I don't need it now so I will remove that. 

Get_dynamic_value didn't return all the attributes of the variable such as type so that's why I added this. I actually don't need get_variable personally and get_variables is really for internal use of supplying the list of variables to companion. But we can have both publicly accessible.

chalamministries:
I added some timer actions to the api

PR #1820, opened 2025-07-07
implemented option to customise the auto font sizing delay

atomicsam:
The auto font sizing delay was too much for our workflow so I have added an option to customise it within the transition menu as other workflows might not care about it. However this may a little too confusing for some users so it may be better to keep this hidden within advanced/other settings.

Also let me know if my implementation is a little inefficient, while this does work I do believe adding another option within the other settings tab and updating the value in the autoSizeDelay store would suffice.

Screenshot_20250707_201557

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2024,20 +2024,6 @@
         "win32"
       ]
     },
-    "node_modules/@rollup/rollup-win32-x64-msvc": {
-      "version": "4.44.1",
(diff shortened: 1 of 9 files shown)


vassbo:
Thanks, but the delay is really a quick fix to prevent showing the text rescaling multiple times to find the correct font size. We should do this in a more future proof way.

We should either find a way to prevent the text from updating while finding the font size, or instead of the delay we start the animation right when the auto size is finished.

Issue #1736, opened 2025-06-15
Collections are not available in RemoteShow


gladsonsam:
**Describe the feature**
Apologies if this exists, but I cannot find a way to control scriptures remotely when we are using a collection (multiple translations/languages). Can we add them to the RemoteShow so that the scriptures column in RemoteShow and FreeShow would be the same.

![Image](https://github.com/user-attachments/assets/d5ecc6ac-6269-4722-b477-8a62d7a5a32c)

![Image](https://github.com/user-attachments/assets/1ba6bfd8-e631-4f16-b1cb-d6bd9e556a85)

gladsonsam:
Also, there is no real way to swap the language/translation shown in FreeShow (for the operator) when using a collection. To my knowledge, it just uses the first one that was selected when making the collection. 

Wasn't a major issue until our operator needed the english translation shown, but we figured it out and made a new collection.

gladsonsam:
Has the feature to swap between the translations in a collection been approved/made? As in if I have a bilingual collection, I can choose which one the operator sees?

vassbo:
Not yet.

PR #1836, opened 2025-07-16
Feature/translation swapping

gladsonsam:
Just to address this (https://github.com/ChurchApps/FreeShow/issues/1736#issuecomment-2989559797):

Allows operators to swap between translations in scripture collections using a swap button in the bottom toolbar. Updates both
the reference label and the displayed verse content. Working concept, might need some polishing such as preserving the translation shown


vassbo:
Cool! I think this can work, we should probably save the chosen preview in the scriptures collection store. Also I'm thinking instead of the "swap" icon button, we can have the button on the actual version name in the bottom left of the drawer (right under all the book names).

PR #1840, opened 2025-07-17
Feature/chord sheet export

gladsonsam:
It works but could use some work. I couldn't get the modal looking nice with the preview. I might try to completely redo it.

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -870,10 +870,7 @@
         "line_spacing": "Line Spacing",
         "metadata_display": "Metadata Display",
         "artist": "Artist",
-        "key": "Key",
-        "song_key": "Key",
-        "chords_used": "Chords Used",
-        "notes": "Notes"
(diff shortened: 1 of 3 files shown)


gladsonsam:
Okay, cleaned up the code now. Just a few questions/things to do:

1. Should I grey out toggles for artist, key and notes if they are blank? 
image

2. Need some help with getting the modal to look good. I'll try make it better tomorrow but preview isn't readable and it has unnecessary scrolling.

3. Title font size?

vassbo:
1. I think we should always print the available metadata, so we don't need "Artist" or "Key". I guess the same goes for notes as well.
2. I can improve it eventually.
3. Does probably not need an input.

vassbo:
Merging for now, we can make the changes eventually.

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.4.8-beta.2

PR #1886, opened 2025-07-28
Startup time

jzongker:
Drastically improved splash screen loading times by loading servers asynchronously, and by lazy loading UI components when needed instead of upfront.

**Production:**
Before: 9.5 seconds
After: 780ms

**Dev:**
Before: 42 seconds
After: 33 seconds

### DIFF
--- a/src/electron/index.ts
+++ b/src/electron/index.ts
@@ -30,7 +30,7 @@ if (!isProd) process.env.ELECTRON_DISABLE_SECURITY_WARNINGS = "true"
 
 // development settings
 export const OUTPUT_CONSOLE = false
-const RECORD_STARTUP_TIME = false
+const RECORD_STARTUP_TIME = true
 
 // get os platform
 export const isWindows: boolean = process.platform === "win32"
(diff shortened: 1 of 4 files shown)


vassbo:
What would be the benefit of the lazy load, as that does not reduce startup time?

vassbo:
I guess the speed a little bit better. The pages were just blank for a few seconds on initial load, but it seems to not be as big of an issue on the prod build.

jzongker:
That second or so delay you saw when trying to load screens was initially occurring when trying to load MainLayout, due to those screens being imported there.  In release, that second or two is reduced to maybe 300ms, and not really noticeable.  However, 300ms times 16 screens still adds up if you try to load them all at launch.  During the average user session, most of these screens won't be accessed, so I think it's better to take the 300ms hit when they are than 300ms*16 hit on startup.

There is a third option available of asynchronously pre-loading these screens in the background after the app launches that we could use if that delay when navigating is too much.

vassbo:
Makes sense, thanks. I think it will be fine.

Issue #1921, opened 2025-08-05
Conditions in StageShow


gladsonsam:
**Describe the bug**
This is my first time using conditions so I could be doing something wrong, but no matter what I try, I can't get a text box to disappear when I visit the StageShow web interface. The preview on the sidebar shows that the text is gone (which is what I want), but when I look at the stage display, the text is still there
**Reproducing**
I just put a random condition: 

<img width="1979" height="368" alt="Image" src="https://github.com/user-attachments/assets/5bfae82e-4b55-4c8c-b1f2-d6df3e522778" />

The left sidebar shows the text is gone:

<img width="2538" height="1362" alt="Image" src="https://github.com/user-attachments/assets/4780d7ed-cb26-4fbc-8c05-fdb16ee08ea4" />

But its still there when I visit stageshow:

<img width="1934" height="1196" alt="Image" src="https://github.com/user-attachments/assets/cea3d7f5-0b80-4c7c-9b37-80c67112eadd" />

**Version (Optional)**
 - OS: Windows 11
 - FreeShow Version: 1.4.8

vassbo:
It only works in the normal stage display, it has not been added to the web interface yet.

gladsonsam:
Okay, is it possible to add that feature then?

Also can we add show_notes as a dynamic value. I'm trying to get my stage display to only show the transliteration if its a song in a different language. My idea was to make a conditional text box that shows if the show is bilingual. 

My first thought was to make a condition if the Text has value ":ml" (also tried "[#1:ml]"), but they don't seem to work (I think the condition only looks at the displayed text, not the hidden text). So my other thought was to put "Malayalam" (or any text) as the show note. then if Dynamic Value show_notes has Malayalam, the text box would show. Hope that makes sense.

vassbo:
Yes, eventually.

It only reads the actual text value yes.
I guess you are looking for "layout_notes", which is already there! Or you can ues "slide_notes", or a custom metadata value.

vassbo:
I recommend just changing the "Item number" to e.g. "1" that should only show the second item if that is a different language. (If you have one show with multiple languages)
<img width="339" height="264" alt="Image" src="https://github.com/user-attachments/assets/7321dbd8-5ddb-4342-922a-7d9a99e19f88" />

gladsonsam:
[...]

PR #1923, opened 2025-08-05
feat: implement stage display condition parity between desktop and we…

gladsonsam:
Should fix https://github.com/ChurchApps/FreeShow/issues/1921

Not sure if this is the best way to indicate why text box is blank:
image

It works for me, haven't thoroughly tested.

- Add comprehensive condition evaluation logic to both frontend and server-side Stagebox components
- Support all condition element types: Dynamic value, Text, Timer, Variable
- Implement case-insensitive string comparisons for consistent behavior
- Add metadata key lookup with case-insensitive matching for dynamic values
- Synchronize condition hiding behavior across desktop app and StageShow web interface
- Add proper variable and timer state checking for condition evaluation
- Remove duplicate removeConditions function declaration

Resolves condition display inconsistencies between physical displays and web interfaces.

### DIFF
--- a/src/frontend/components/stage/Stagebox.svelte
+++ b/src/frontend/components/stage/Stagebox.svelte
@@ -1,7 +1,7 @@
 <script lang="ts">
     import { onDestroy } from "svelte"
     import type { StageItem, StageLayout } from "../../../types/Stage"
-    import { activePopup, activeStage, activeTimers, allOutputs, currentWindow, dictionary, outputs, outputSlideCache, previewBuffers, refreshEditSlide, stageShows, timers, variables } from "../../stores"
+    import { activePopup, activeStage, activeTimers, allOutputs, currentWindow, dictionary, outputs, outputSlideCache, previewBuffers, refreshEditSlide, showsCache, stageShows, timers, variables } from "../../stores"
     import { sendBackgroundToStage } from "../../utils/stageTalk"
     import EditboxLines from "../edit/editbox/EditboxLines.svelte"
     import autosize from "../edit/scripts/autosize"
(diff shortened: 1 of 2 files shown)


vassbo:
If the condition is not met the item should not be visible at all! (In the editor it should always be visible, just like it already is - there can be an indicator, maybe the light bulb can change color when the condition is active)

You can copy pretty much everything directly over from how it is in the stage display output.

vassbo:
The changes you made in "/frontend/components/stage/Stagebox.svelte" can probably be reverted, as it is duplicate code.

gladsonsam:
Whoops, I'll clean that up. The item isn't visible on stage displays if the conditions aren't met. I'll change it so that in the editor, it will still display the text, but the lightbulb will change colour when the condition is met.

vassbo:
The logic in `StageSlide.svelte` (frontend) for the condition can be copied over to the `Slide.svelte` (stage), instead of having the logic in `Stagebox.svelte`. (Or the way you have done it works, but we should extract the functions into a different file)

And we can probably request the value of `isConditionMet` function in stage from the frontend, to get all the nessesary values?! Depending on how easy it is to get the proper values to the stage.

vassbo:
I just did it myself, it was pretty much copy paste. Some values like dynamic values does not work at the moment.

https://github.com/ChurchApps/FreeShow/commit/1e7aaa10400257757515fcfe5124969efce8ff42

vassbo:
You can try to fix the things that currently does not work if you want to, also the stage editor preview is not added yet.

gladsonsam:
Tried to fix things, but for some reason, my conditions are doing the opposite of what they should be doing. I think its something on my end as it was working before. Will submit a PR if I can get there.

PR #1944, opened 2025-08-08
Feature/stage editor copy paste

gladsonsam:
This adds the ability to shift select multiple items in Stage editor, and copy and paste items.

I haven't been able to get dragging multiple items working. Rectangle select would also be nice, so it feels similar to PowerPoint etc.

### DIFF
--- a/src/frontend/components/edit/editbox/Editbox.svelte
+++ b/src/frontend/components/edit/editbox/Editbox.svelte
@@ -41,30 +41,6 @@
                 return ae
             }
 
-            if (e.shiftKey && ae.items.length > 0) {
-                // Shift selection: select range between last selected and current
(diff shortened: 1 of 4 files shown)


vassbo:
You can already select multiple items with the ctrl key, so all the integrations you did with the shift key is not nessesary as they can just be used with the existing code if shift is better than ctrl.

The changes you made to clipboard for the copy/paste is good though.

gladsonsam:
Didn't see that. To me, shift makes more sense, as that's how most things multi select.

vassbo:
Yeah, we can probably change it to shift if that's more common.

gladsonsam:
I think it's good now? Let me know if more changes are needed

Edit: forgot about changing to shift, will do that

PR #1952, opened 2025-08-11
Feature/ultimate guitar search

gladsonsam:
Just an idea I had, but it worked out. It does use another [package](https://www.npmjs.com/package/ultimate-guitar), which doesn't have many downloads, but it seems to work fine. 

https://github.com/user-attachments/assets/530d1753-4498-48db-9748-d5f6f75ad7d7

If need be, I can improve the parser in Quick Lyrics.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
     "name": "freeshow",
-    "version": "1.4.8",
+    "version": "1.4.9-beta.1",
     "lockfileVersion": 3,
     "requires": true,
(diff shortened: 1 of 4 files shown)


vassbo:
Nice, we might have to do something about it pulling the same title/artist many times (I guess there might be some variations, like the chords being different). But this seems good for now.

gladsonsam:
True didn't think about that. I'll be using this feature a lot, so I'll find any issues I have with it.

Issue #1637, opened 2025-05-17
MacMini M4 - Local Network Name Already in Use error


Mark-SoundEngineer:
Whenever 'RemoteShow' or 'StageShow' are enabled in Settings: Connections, the 'Network Name is Already in Use' error appears when the Freeshow is started. Every time.

Turning off these two options, prevents the error. After the error is accepted your mac local network make is changed (higher number added).

MacMini M4 Sequoia
 - FreeShow Version: V1.4.2 and V1.4.3

'RemoteShow' or 'StageShow' still work, but the operator has to clear the error each time the app is opened. The issues on different networks, but only on a Mac. A PC on the same network does not have the same issue.

Ideas??

Mark-SoundEngineer:
![Image](https://github.com/user-attachments/assets/b8fa3f30-ceb0-4b15-8789-ecbbc2bb131b)

armm29393:
Same with Macbook Air M2

vassbo:
That has happen to many mac users lately.

@Auxority Do you think it's caused by a package update?: https://github.com/ChurchApps/FreeShow/pull/1529

vassbo:
Got it. But it did not happen previously, only the last few weeks.

andrewgribben:
Yes it's definitely a recent occurence. I feel like there's some interplay between NDI, Bonjour and how the computer is advertised on the localnetwork

Auxority:
@vassbo we can always try to revert it locally and see what happens.

vassbo:
I don't have a mac, so I can't test it.

Auxority:
Ahh alright, then I’ll try later this week.

Mark-SoundEngineer:
> Have you tried to change the port number? Maybe add a 1 to the end of the port number, so 5510 becomes 55101.
> Gear icon->Connection->RemoteShow->Gear Icon->Port Gear icon->Connection->StageShow->Gear Icon->Port

I tried the above suggestion, but no joy I'm afraid.

vassbo:
@Mark-SoundEngineer What if you go back to FreeShow version 1.4.0, does that version not have this issue?

Auxority:
I experienced the same issue on 1.4.3 

![Image](https://github.com/user-attachments/assets/ffde5521-3064-4c4b-bcd8-3a7d11889b26)

vassbo:
[...]

PR #1958, opened 2025-08-11
upgrade minor versions of npm dependencies

Auxority:
Just ran these commands:

```sh
npx npm-check-updates --target minor -u
npm install
```

Hopefully fixes this: https://github.com/ChurchApps/FreeShow/issues/1637

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -43,7 +43,7 @@
                 "socket.io": "^4.8.1",
                 "socket.io-client": "^4.8.1",
                 "sqlite-to-json": "^0.1.3",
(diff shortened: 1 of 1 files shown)


Auxority:
image

Not sure whether I should include this changed PDF.

vassbo:
Remember that `sqlite3` must stay at "5.1.6" still!

The pdf worker can be changed as that updates with the new `pdfjs-dist`  version.

Auxority:
I'll lock that version in the `package.json` then, to prevent issues in the future. Thanks for the reminder!

Perfect :)

PR #1975, opened 2025-08-14
1.4.9-beta.2

vassbo:


linuxnow:
Images don't load in this pre-release but do work in beta1.
I'm using Fedora 42 x86_64s provided rpm.

vassbo:
Thanks, I assume they don't load in the media drawer?

@Auxority Seems related to some changes you made? https://github.com/ChurchApps/FreeShow/pull/1965

Auxority:
Gotta love filesystems, it’s probably because it generates the paths differently now. It might work if he adds new media.

But gotta brainstorm a bit on a good solution here. Maybe we should try the old filepath first, and use the new method as a fallback. I’ll try to create a PR today.

linuxnow:
> Thanks, I assume they don't load in the media drawer?
> 
> @Auxority Seems related to some changes you made? #1965

They don't load in the output and I have problems loading some in the media drawer. It's kind of strange.

vassbo:
@linuxnow https://github.com/ChurchApps/FreeShow/pull/1985

Auxority:
If the new solution does not work, we'll have to look into adding a fallback. Didn't want to do that as a first step though. Maybe we could just generate a "srcset" instead of a "src". That way we can add multiple paths.

vassbo:
I guess. But it seemed to work fine in 1.4.8 in both scenarios, so there is a change that broke it I guess.

Issue #1982, opened 2025-08-15
UI Bible Feedback


MyGuyConsultingYou:
**Describe the feature**
Love the new look in the 1.4.9 beta-2. 
It would be helpful to see the titles of the 3 Bible categories labeled above each of their respective sections.
I've gotten used to the groupings but for new or beginner users it may prove helpful especially when trying to explain how to do things to volunteers.

<img width="649" height="634" alt="Image" src="https://github.com/user-attachments/assets/63d9c820-ad9d-4757-ae7f-040115ce2006" />

gladsonsam:
Does this work?

https://github.com/user-attachments/assets/c2092f91-6670-4774-9aa1-f5ee26df22fd

vassbo:
<img width="351" height="685" alt="Image" src="https://github.com/user-attachments/assets/84be20f4-0ed1-480a-acba-2d8485d892f7" />

Also titles for other sections:
<img width="350" height="458" alt="Image" src="https://github.com/user-attachments/assets/554c0cab-bbc1-4397-b346-db5518f5cae7" />

vassbo:
Creating a Bible Collection should also be more intuitive:
<img width="800" height="626" alt="Image" src="https://github.com/user-attachments/assets/1e6b047d-af15-4159-bee1-15537b7bba63" />

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.4.9-beta.3

PR #1984, opened 2025-08-15
Feature/UI scripture updates

gladsonsam:
Addressed (https://github.com/ChurchApps/FreeShow/issues/1982) by adding labels
Also adds a new collection button so its easier to create collections

I also want to improve the shift click as currently you have to shift click on a scripture, even if it is active.

https://github.com/user-attachments/assets/32acf516-9ace-4248-bcc6-bdbbae9395ee

I think it's better that the active scripture is already part of the selection, so you just need to shift+click on the other scriptures to add to collection.

Let me know what you think


vassbo:
I did consider that this might be a good idea, thanks for the PR. I think the title sections should be much smaller. Also I don't think there's a reason to seperate Local VS API Bibles as they are pretty much the same thing for the end user, so we can keep them in the same section as before and have one "Bibles" title.

vassbo:
The reason it has not been done it because the drawer tabs itself is a title, but now that it's more divided up into sections it's probably good to have titles.

vassbo:
Maybe also have subtitles under the title that say "Local" and "API", but they should be in the same section because of the "New scripture" button.

gladsonsam:
Alright do you still want the line divider between local and api though?

vassbo:
Yes, as they can't be selected at the same time (for now).

thebiblelover7:
I think there should be some distinction between local and API Bibles as they are sometimes confusing to the end user when they can't be grouped together.

gladsonsam:
Struggling to find a good way to distinguish local and API. This is the best I could come up with:
image

let me know if this works or I might just leave it out for this PR

vassbo:
Nice. I think that can work, maybe we can move them over to the left side and add a bit more opacity.

gladsonsam:
Whoops ignore that typo in my commit (I meant shift). I have got shift click to select all the ones in between and cntrl click to select specific ones. Shift click is a bit buggy on the API one but I can't seem to fix it.

vassbo:
No worries. 😅 Nice, that selection code was very outdated, so glad to see some improvements.

vassbo:
Did you test that ctrl/shift selecting slides still works as expected?

vassbo:
I see that it's specific for `category_scripture` which works for now, but there should be a more universal system, but it can wait for now.

gladsonsam:
Yeah, I'll try work on a better selectelem if I can

vassbo:
The collection system is not good. This is more of what I had in mind: https://github.com/ChurchApps/FreeShow/commit/52c2549be3b9cd9ad612be7814a116a98c290d1a

image

image

vassbo:
I think a popup select is better than selecting the tabs, and also much more intuitive than the previous select right click.

PR #1985, opened 2025-08-15
fix filepaths for all operating systems

Auxority:
Tested on (Arch) Linux and MacOS. Both the media drawer and the choose media option while editing templates work fine.

I do get this error in the logs, even though the file exists. But the media still plays just fine.

```
MP4Box error catch: Error: ENOENT: no such file or directory, open 'file:///home/pengu/Videos/test.mp4'
    at Object.openSync (node:fs:573:18)
    at Object.func [as openSync] (node:electron/js2c/node_init:2:2559)
    at Object.readFileSync (node:fs:452:35)
    at t.readFileSync (node:electron/js2c/node_init:2:10175)
    at /home/pengu/Documents/code/svelte/FreeShow/build/electron/utils/files.js:527:59
    at new Promise (<anonymous>)
    at /home/pengu/Documents/code/svelte/FreeShow/build/electron/utils/files.js:524:16
    at Generator.next (<anonymous>)
    at /home/pengu/Documents/code/svelte/FreeShow/build/electron/utils/files.js:33:71
    at new Promise (<anonymous>) {
  errno: -2,
  code: 'ENOENT',
  syscall: 'open',
  path: 'file:///home/pengu/Videos/test.mp4'
}
```

### DIFF
--- a/src/frontend/components/helpers/media.ts
+++ b/src/frontend/components/helpers/media.ts
@@ -65,26 +65,29 @@ export function joinPath(path: string[]): string {
 export function encodeFilePath(path: string): string {
     if (typeof path !== "string") return ""
 
-    // already encoded or special URL
+    // Skip if already encoded or is a special URL
     if (path.match(/%\d+/g) || path.includes("http") || path.includes("data:")) return path
 
-    // can't load file paths with "#"
-    // Ensure proper file:// protocol for absolute paths on Linux
-    if (path.startsWith("/") && !path.startsWith("file://")) {
-        path = `file://${path}`
-    }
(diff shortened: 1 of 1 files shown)


Auxority:
If I leave out the `file://` prefix, media used in templates does not work. But if I add the `file://` prefix, I see that error in the logs, even though the media loads fine.

vassbo:
I guess nodejs does not support `file://` while it's needed in the frontend.

In files.ts the `extractCodecInfo` function probably can be updated to trim the `file://` and that should probably fix it.

Auxority:
Ohh, that’s probably it!

PR #2027, opened 2025-08-24
Feat/texteditor find and replace

gladsonsam:
This is more of what I had in mind for a find and replace for the text editor, but see what you think. The find panel might need a better UI. Let me know if any changes should be made


vassbo:
Not sure if we need the searching? Also this implementation is very buggy. The reason for the popup is so that can be used universally (the current popup UI should be updated at least).

PR #2040, opened 2025-08-26
Add arm64 snap builds

eeickmeyer:
While official support is up to you, this enables arm64 builds of the snap, which I was successful in building locally.  Successfully built and tested on a Raspberry Pi 5. Can be further tested on the `latest/edge` channel on an arm64 device.


vassbo:
Cool, I don't mind extra arm support! :)

PR #2064, opened 2025-09-02
Feature/add test suite for the application

otonielpv:
This PR introduces a comprehensive test suite for the entire application. Its purpose is to ensure that any new feature or change implemented does not break the current functionality of the application. This will help prevent functionality-related issues, such as features that previously worked but no longer do.

For more details on what has been included, please refer to TESTING_SUITE_SUMMARY.md or tests/README.md.

If you would like to add more tests, feel free to let me know, and I will be happy to assist. Any type of feedback is also highly appreciated—please don’t just close the PR without sharing your thoughts 😄.


vassbo:
This seems nice, but it looks like most of the tests & code is redundant!

vassbo:
I also think it's too much to maintain.

otonielpv:
@vassbo Well, I think the tests suite are one of the most important things you should have on your software, why you say it is too much to maintain, the tests are covering the whole functionality of the application, I think you won't redo the whole application now, so what this is doing is to ensure that nothing that is being added after the tests will brake nothing of the functionality that is already on the application, I think this is specially important on open-source softwares. What do you think?

vassbo:
It's helpful to have tests, yes, but a bunch of the tests here were checking unimportant values like if the value is null or undefined. And other checks I saw that were redundant. Also at the moment a lot of the code has been reworked to be cleaner. In my experience the things that break is rarely the common things.

otonielpv:
I deleted the redundant tests and also the e2e (you can check it here https://github.com/otonielpv/FreeShow/commit/ad0abefa1b53426ad2bc8bfa07d40fefc32bd2fc), I think the tests now are something really helpful, they take less than 10 seconds to execute and they are insurance in case someone adds something they shouldn't. As I told you if you have copilot it is really easy to add test for the new functionality that you add

vassbo:
I still se a bunch of redundant tests, here's just one example, that is not helpful:

```
outputState.background = slideOutput
expect(outputState.background).toEqual(slideOutput)
```

otonielpv:
Well reviewing It I see that It is really hard to test the code without mocls so It is not worth

PR #2069, opened 2025-09-02
Chord sheet option appears immediately when entering PDF export modal

gladsonsam:
Currently, the chord sheet option only appears after navigating back from the PDF modal to the format selection and then back to PDF again.

Pretty sure this is a bug, so this PR should fix it.

### DIFF
--- a/src/frontend/components/main/popups/export/Export.svelte
+++ b/src/frontend/components/main/popups/export/Export.svelte
@@ -155,7 +155,7 @@
 
     {#if exportFormat === "pdf"}
         <HRule />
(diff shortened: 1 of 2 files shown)


vassbo:
You are correct, but the way you fixed it seems overcomplicated.

vassbo:
This should do the trick: https://github.com/ChurchApps/FreeShow/commit/65ec32b16f7017ab622bfeecd9c682215366cc1c

gladsonsam:
One day, I'll get to your level 😅

Thanks

PR #2078, opened 2025-09-04
Allow snap to build on amd64 still

eeickmeyer:
I'll do you one better. :)


vassbo:
Nice. Maybe also skip `npm run replace-electron` on amd.

vassbo:
It did still work on amd with the pack:arm build, but this is probably better.

eeickmeyer:
> Nice. Maybe also skip `npm run replace-electron` on amd.

Done.

PR #2084, opened 2025-09-05
Start button not working on clean installation

otonielpv:
So I Installed FreeShow on a new PC for the first time and the start button was not working, I had to close and save and then start again the program


vassbo:
Thanks. This is important! I suppose this only happened in the beta. I don't even know why I added that check.

PR #2101, opened 2025-09-10
Added feature "First Day of Week" as regional setting

mayrsau:
Here is a small contribution to this awesome project:  Adding a setting for adjusting the first day of the week for calendar and event scheduling purposes.   This allows the application to be configured to match regional settings, and helps planning full-week events.

Also upgraded language files dependent on this feature, and fixed several typos to enable searching code better.


vassbo:
Please remove the typo fixes in this PR as it makes it hard to see what's new. Also some of the typos must stay as the user data depends on it. (Also the language files should not be updated here as they depend on data from Transifex)

vassbo:
Btw. thanks for the PR, :) I'm thinking this should be an option in the "Calendar" page, instead of the general settings, as it's only used there. Also if we can auto get the preferred value that would be great.

vassbo:
If you would like to translate, please join Transifex: https://explore.transifex.com/nettbiter/freeshow/

vassbo:
This is what I was thinking: https://github.com/ChurchApps/FreeShow/commit/b4abb027616f35a55e4a45f0155a0fb00e64c101

I also used the `new Intl.Locale(languageId).getWeekInfo().firstDay` to automatically get the preferred first day.

mayrsau:
> This is what I was thinking: [b4abb02](https://github.com/ChurchApps/FreeShow/commit/b4abb027616f35a55e4a45f0155a0fb00e64c101)
> 
> I also used the `new Intl.Locale(languageId).getWeekInfo().firstDay` to automatically get the preferred first day.

Thanks, Kristoffer.  This resolved our issue.  My only concern is finding those settings.  It is not obvious to a casual user.  I will add it to the documentation.

vassbo:
Thanks. I'm thinking it does not need to be very visible as it should set it automatically.

mayrsau:
> Thanks. I'm thinking it does not need to be very visible as it should set it automatically.

Yes, I think this would be acceptable the way it is if the settings actually pick up from the locale in my host OS (Windows).  In my case, it is not picking it up, so having that settings option to override it solves my immediate problem.  

I am not sure where `Intl.Locale(languageId).getWeekInfo().firstDay` is getting its data from.  I see the `FreeShow/node_modules/dayjs/esm/locale` settings are not coming through, and my OS locale either.  I have my language set to `en` (FreeShow & OS) and my regional settings `United States` (OS).

Another issue I see with the automatic setting: currently the `Calendar.svelte` file seems to support only Sunday & Monday as first day of week (the most common cases), but in some locales, they use Saturday instead (and my OS lets me actually set it to any day of the week), which would break the logic if set automatically to Saturday.  Correct me if I am wrong, since I did not review the code in detail.

Here is a map with these preferences:
[First Day of Week Map](https://en.wikipedia.org/wiki/Week#/media/File:First_Day_of_Week_World_Map.svg)

vassbo:
It's getting data directly from the computer. It should not break, it can do for now. You can check it out in the next version.

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.5.0-beta.2

PR #2137, opened 2025-09-15
Enable backdrop filters on stage editor

otonielpv:
So I needed to blur the background of the textbox from my streaming stage layout and I found that the backdrop filters where disabled for the stage editor, there was any reason to have them disabled? I also fixed an error with the autosize that was giving on the build of the app


vassbo:
Nice, thanks. The reason it was disabled was because I would not think it was needed on the stage display.

PR #2141, opened 2025-09-16
Enable cams on startup and improve cameras performance

otonielpv:
So I have a stage output to control all my streaming output, I created multiples stage layouts for different moments on the service, for example for worship (camera + slide texts), for images (camera + images), etc. I was having a lot of issues making the camera to work, because sometimes it worked only in few layouts and not in others. I used the trick of an overlay to start the camera but It didn't work, with this the cameras will start on the startup of the program and now they load waaaay faster.


otonielpv:
@Auxority you will be surprised when I tell you that this has been made by copilot 😝

vassbo:
Copilot tends to overcomplicate things, I got it improved and reduced to under 50% of the code.
I changed the activate on startup to be a context menu action in the camera drawer.
The camera manager is nice to have, so I also implemented that globally.

PR #2153, opened 2025-09-18
Improve current output for stage layout

otonielpv:
I use a lot the current output on the stage layout for my streaming, here is a feature to enable or disable the direct rendering, with direct renderingthe performance is better, here is a video showing the performance enabling and disabling the direct rendering, by default all the current output have the direct output disabled

https://github.com/user-attachments/assets/b26a3ba3-18ba-4d14-958c-85c74e117d5c

### DIFF
--- a/src/frontend/components/stage/Stagebox.svelte
+++ b/src/frontend/components/stage/Stagebox.svelte
@@ -266,8 +266,8 @@
         <span style="pointer-events: none;width: 100%;height: 100%;">
             {#if item.type === "current_output" || id.includes("current_output")}
                 {#if !$special.optimizedMode}
-                    <!-- Use PreviewCanvas when in output window (stage projection) or preview mode -->
-                    {#if $currentWindow === "output" || preview}
+                    <!-- Use PreviewCanvas only in output window (stage projection) -->
+                    {#if $currentWindow === "output"}
                         <PreviewCanvas capture={$previewBuffers[stageOutputId]} id={stageOutputId} fullscreen />
                     {:else}
                         <Output outputId={stageOutputId} mirror preview={preview} style="width: 100%; height: 100%;" />


vassbo:
Why not use a normal output? The reason for the capture is because of the StageShow, I don't see a reason to use that when you can just use a normal second output!

otonielpv:
I tried to use the normal output but I found the stage one more easy to use and It gave me more flixibility than using normal output and styles/templates

otonielpv:
Even though I use It for streaming I think this may help people who uses current output for another things

vassbo:
I see.. We should not have a toggle, but just use the output component directly always in the main app/output then.
Just make sure the StageShow "Current output" still works remotely!

vassbo:
If you can figure it out, we also don't need to enable the output capture if there is no stageshow connections. in outputs.ts>shouldBeCaptured

otonielpv:
So the current output is working remotely through the StageShow

https://github.com/user-attachments/assets/b9b12460-dfb6-4627-99ef-925070506bdc


I did not understand this:

> If you can figure it out, we also don't need to enable the output capture if there is no stageshow connections. in outputs.ts>shouldBeCaptured

What do you mean?

vassbo:
That would save even more performance as it would not start capturing the output window when it's not used. I can try to do it. But is it better with your last changes? As it seem to always use the canvas and capture now. But I guess the throttle changes to PreviewCanvas helped a lot.

otonielpv:
Sorry but I don’t get it 🥲, if you detail me the problem I could ask copilot to do it.

vassbo:
It's fine. Is the performance still good? Looks like the canvas was causing the trouble, actually. As now the canvas is set to be active in both your output, and preview: `$currentWindow === "output" || preview`

otonielpv:
Yeah I noticed on the preview that It hasn't the same performance as the real stage output, now it's fixed the canva is only used on the output window

vassbo:
Nice, thanks.

PR #2255, opened 2025-10-21
added toggle output button

gladsonsam:
Hope this doesn't break anything, but this adds the toggle output to RemoteShow

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -76,8 +76,9 @@ jobs:
           sudo apt-get update
           sudo apt-get install -y libfontconfig1-dev uuid-dev --fix-missing
 
-      - name: Use stock Electron on ARM64
-        run: npm run replace-electron
+      # WIDEVINE
+      # - name: Use stock Electron on ARM64
+      #   run: npm run replace-electron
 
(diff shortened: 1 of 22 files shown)


vassbo:
Why do you need this? The main problem is that you can accidentally disable the output windows.

I would recommend always having the output windows active, also you can already bind actions to slides to toggle the output remotely if you want to.

vassbo:
Also the remotes should be kept relatively simple with not too many buttons, as that may be confusing.
But you are welcome to convince about a good reason for this.

gladsonsam:
We often use FreeShow alongside other apps (etc, PowerPoint), so I often use the toggle display setting to project other apps. I usually use RemoteShow because I am not at the lyrics computer, so that's why I added this. However, if it's not needed, I'll just use the API command. 

I understand that RemoteShow should be relatively simple, but I think an exception can be made for tablet/desktop modes, where it would be nice to extend the features to include actions, etc.

Just my two cents. Let me know what you think.

vassbo:
Right, but I think it's not needed for most people.
You can use the API command, or automate it with actions when you run a specific slide from RemoteShow.

Issue #2316, opened 2025-11-03
RemoteShow font size in Safari


hpskk7wg7s-sys:
In the FreeShow Remote App, it displays all slides nicely on Android, while on Apple it displays them enlarged.

gladsonsam:
The issue has something to do with mobile and desktop views, where only mobile views display the slide text properly.

vassbo:
Use the Chrome browser, it's an issue with Safari: https://github.com/ChurchApps/FreeShow/issues/1167#issuecomment-2783769723

Or turn off auto size.

vassbo:
@gladsonsam I guess we can't change the native webview on iOS to not use Safari.

gladsonsam:
I will take a look today. I guess a quick fix could be to request desktop mode with userAgent. Will experiment today and see.

PR #2320, opened 2025-11-04
Replaced CSS zoom with transform-based scaling

gladsonsam:
It works, but I'm not sure if it's the ideal way. I couldn't get userAgent working in FreeShow Remote, and I think the problem is best solved from the root.

https://github.com/ChurchApps/FreeShow/issues/2316

Tested on mobile, desktop and iOS


vassbo:
Cool, thanks! I guess this is the best way for now. It's strange that Firefox & Safari doesn't support the `zoom` key properly.
https://caniuse.com/css-zoom

PR #2335, opened 2025-11-07
Add additional snapcraft plugs

eeickmeyer:
I noticed the snap wasn't broadcasting NDI. This should fix that, and add GPU hardware acceleration access.


vassbo:
Cool, thanks!

PR #2336, opened 2025-11-08
Fix RemoteShow Scripture Navigation and Highlighting

gladsonsam:
- Verse highlighting now syncs
- Search navigation works properly
- Code cleanup

Let me know if any changes are needed.

### DIFF
--- a/src/frontend/utils/remoteTalk.ts
+++ b/src/frontend/utils/remoteTalk.ts
@@ -169,25 +169,35 @@ export const receiveREMOTE: any = {
         }
 
         if (bookKey && !chapterKey) {
-            const bookData = await jsonBible.getBook(bookKey)
-            const mapped = (bookData.data.chapters || []).map((c) => ({
-                number: c.number,
-                keyName: c.number,
-            }))
-            msg.data.bibleUpdate = { kind: "chapters", id, bookIndex, chapters: mapped }
+            try {
+                const bookData = await jsonBible.getBook(bookKey)
+                const mapped = (bookData.data.chapters || []).map((c) => ({
+                    number: c.number,
+                    keyName: c.number,
+                }))
+                msg.data.bibleUpdate = { kind: "chapters", id, bookIndex, chapters: mapped }
+            } catch (error) {
+                console.warn(`Failed to load book ${bookKey}:`, error)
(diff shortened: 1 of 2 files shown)


gladsonsam:
Forgot about API bibles, will get that working then submit for review

vassbo:
Thank you!

gladsonsam:
I have another problem with remoteshow that I am struggling to come with a solution.

If we have a list of shows in the projects, our slide operator can navigate them, and we can transpose as long as the selected show is the show we want.

If I decide to play a show from the remoteshow, I can't transpose the show because the show isn't selected. The operator can't see what show is being played either (I think), so it's not like they can quickly select it.

My current solution is to add to the project and then remote in and select the show so we can transpose it, so everyone gets it in the right key.

This might be one of those problems that only I have, so I'll let you decide what a good solution is.

I'm happy to make a PR

vassbo:
Okay, so making transpose work when the show is not active would fix this?

gladsonsam:
Yeah, but what would it update. Maybe the show of the current slide? That could work.

vassbo:
Oh, I get what you mean I think.

In the right side under the preview and clear buttons, the operator can see the currently outputted show, and click on the title to open it.
If we implement that into the remote as well you can click on the title to open it!

image

vassbo:
So that means the transpose doesn't need to change.

gladsonsam:
> In the right side under the preview and clear buttons, the operator can see the currently outputted show, and click on the title to open it.
> If we implement that into the remote as well you can click on the title to open it!

I didn't know this, but this is pretty good. I don't think adding this will allow me to transpose on stage until the operator clicks this right?

vassbo:
From RemoteShow I think you can transpose the show you have opened locally, regardless of what's being outputted, and regardless of what's opened in the program.

gladsonsam:
Oh, I didn't see a transpose option in remoteshow, but if that exists thats perfect

gladsonsam:
image

Would something like this work? Two buttons in the text editor, and moving the transpose so it's a shared util

vassbo:
Yeah, that's cool. I though it existed, but remember incorrectly.

Issue #2326, opened 2025-11-06
1.5.2 seems a lot slower to change slides


yurividal:
Hi @vassbo , check this out. There is a big time gap between changing the slide in the control, and it being changed in the output.
The delay was always there, but it seems to have gotten worse on this version?

Not sure if it is related to my template... it is a simple template though... or maybe it is because i have over 3000 songs in my database? I don't know, but it does seem like something needs to be optimized for outputs.

https://github.com/user-attachments/assets/5bc8ada9-6c29-4f1b-87d3-8353f21a1670

yurividal:
It is definetly related to the template. I removed the template overwrite from Settings -> general -> Slides -> Manage groups, and removed all the overwirtes. Then, i set my song to default template, and voila, the slides change IMMEDIATELY.

This is what my Song Template looks like:


<img width="1688" height="945" alt="Image" src="https://github.com/user-attachments/assets/e67af0fb-65d5-4a0d-a216-6ca6aa631856" />

Maybe its the %Progress ??

yurividal:
Yes, it is definetly the %Progress that is causing the massive delay. I removed it from my template, then i re-added all the overwrites, and it is still immediate. So, the delay is definetly caused by the %Progress calculation... Maybe it needs to be calculated in advance, and simply added to the slides as a text?

vassbo:
Thanks for finding the cause. I didn't change anything related to the timing, so shouldn't have been a difference from the beta. I recommend placing the progress item in an overlay, that way it doesn't need to refresh all the time.

vassbo:
I think it's caused by the auto size actually. As it has to recalculate the size everytime you update the slide, so by using overlays you will not have that problem.

yurividal:
Cool, thanks
I'll give the overlay a try and let you know how it goes.

Do you mean the auto-size only inside the box that contains the progress?

Or my verses text box too?

vassbo:
Mainly the text in the progress item, those are calculated seperately, but I think the progress autosize is more expensive than the plain text autosize.

yurividal:
@vassbo the overlay indeed is a lot faster. But I can't seem to find a way to not show the overlay on the OBS output. 
As a text box, i was able to set it to only show on "Main", but i don't see the option to do the same with the overlay

vassbo:
Right click the item, "Specific outputs" same as on a slide.

yurividal:
[...]

PR #2342, opened 2025-11-08
Implement Progress Cache

yurividal:
%Progress data is cached per show/layout. This now reuse layout refs and precomputed group metadata unless the layout or group definitions actually change, and renders using the cached full/one-letter names.


vassbo:
Thank you! Looks nice, but I didn't see a noticable improvement to the speed, did you? Still pretty slow with lots of progress items. We should probably do the cache on a more global level.

https://github.com/ChurchApps/FreeShow/issues/2326

yurividal:
@vassbo I did see a signifficant improvement is speed with this commit. Here is a video of that same song i sent you last week, with this commit. Its immdiate to change slides. (I made the progress green and pink, to make it easier to spot on the test template.

https://github.com/user-attachments/assets/0dcf51b1-fc62-4530-ab15-650ff8576683

Maybe you didn't see improvement because you have auto-size in your progress box? This one in the video is set to fixed size (40) and applied directly to the template - just like last week's test, when it was lagging signficantly

vassbo:
I guess I didn't turn off auto size, you might be correct. Thanks.

PR #2344, opened 2025-11-08
Split verses improvements

yurividal:
This PR solves an issue with "Split Long Verses" 

Currently, if user splits long verses, the resulting slides show the letters (example 2a, 2b etc,..) this is inconsistent with the Bible, and also doesn't match the behaviour seen in "Right-Click -> Format -> Split in Two" 

With this PR, Actual verse numbers don’t carry suffixes, so hiding them keeps the projection closer to scripture—though a toggle remains for teams that still want the letters. It also ensures any click on a split verse grabs every segment, so you can’t accidentally queue only half the passage.

Before:
image

After:
image

It also adds an option that makes it behave as the old way, adding the sufixes, just in case someone likes it...

image


yurividal:
ignore tihs one, it has a small bug.

PR #2345, opened 2025-11-08
Fix "Split Long Verses"

yurividal:
This PR solves an issue with "Split Long Verses" 

Currently, if user splits long verses, the resulting slides show the letters (example 2a, 2b etc,..) this is inconsistent with the Bible, and also doesn't match the behaviour seen in "Right-Click -> Format -> Split in Two" 

With this PR, Actual verse numbers don’t carry suffixes, so hiding them keeps the projection closer to scripture—though a toggle remains for teams that still want the letters. It also ensures any click on a split verse grabs every segment, so you can’t accidentally queue only half the passage.

Before:
image

After:
image

It also adds an option that makes it behave as the old way, adding the suffixes, just in case someone likes it...

image


vassbo:
I think it's common to use a/b to mark parts of a verse. Also that let people know what verse it's a part of. But if you think people don't want that, I guess it's good to have an option to disable it.

yurividal:
Yeah, it makes sense on the preview. I don't think it makes a lot of sense on the actual slide, for the church, since as I said, it's not part of the Bible.

Anyways, yeah, having the option is probably the best outcome. =)

vassbo:
In my area it's common practice to have a/b to mark parts of a verse, but most of the time the full verse is shown. Verse numbers in general is "not part of the Bible", it's just added as a help to see where you can find a certain part.

yurividal:
You are right. When i say " part of the bible" I meant that is something the people can see on the screen that is exactly the same on their bibles. I guess it's just a regional thing, don't think I've ever seen a/b on a verse, except for when someone wants to ONLY refer to a part of a verse
It's what people are used to.. anyways, hope having the option is a good solution.

vassbo:
Yeah I guess `except for when someone wants to ONLY refer to a part of a verse` is when it's used most commonly.
I will merge this PR, but need to make some small changes first.

vassbo:
Instead of +5187 changes it's now +41, making it easier to see the actual changes.

PR #2346, opened 2025-11-08
Fix custom template first slide

yurividal:
Auto-apply  “custom template on first slide” override during template updates and scripture show creation, including background/media handling.

This ensures that the user doesn't need to reaply templates to get the "custom template on first slide" to work.

### DIFF
--- a/src/frontend/components/drawer/bible/scripture.ts
+++ b/src/frontend/components/drawer/bible/scripture.ts
@@ -446,6 +446,7 @@ export function getScriptureSlides({ biblesContent, selectedChapters, selectedVe
 
                 const { id, subverse, endNumber } = getVerseIdParts(v.verseId)
                 const value = `${id}${endNumber ? "-" + endNumber : ""}${subverse ? getVersePartLetter(Number(subverse)) : ""} `
+
                 slideArr.lines[lineIndex].text.push({
                     value,
                     style: verseNumberStyle,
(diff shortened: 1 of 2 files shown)


vassbo:
I tested this with the "Convert to show" scripture button, but the first slide didn't get the "Custom template on first slide" until I clicked on the template. Is that what you meant?

yurividal:
No, on my tests, it gets populated immediately as soon as you do "convert to show", as long as you have "custom template on first slide" configured for the scripture template. 

I'll get a video of it working, for you to take a look.

vassbo:
Okay, thanks. I don't think there's any logic in place to set the custom template on the first slide, as I never implemented that. I tried with this PR, but didn't see any changes to that specifically.

Make sure to use a distinct template, I set the default "Blue Content" with a big blue box on the first slide to see that it doesn't get applied. That's also visible by no icon showing on the first slide. But clicking on my custom scripture template applies the template on the first slide.

yurividal:
https://github.com/user-attachments/assets/be559e46-61bd-4d03-854d-43f37b702625

yurividal:
> I don't think there's any logic in place to set the custom template on the first slide, as I never implemented that. 
>

The button definetly already existed there. it just didn't do anything, before this PR.

See in the video how it works now?

vassbo:
Thanks, I think I just didn't have the correct code after all.

PR #2349, opened 2025-11-09
Better verse split

yurividal:
This commit makes it so that "Split Long Verses" actually uses the same logic as Right-click -> Format -> Split in Two, to implement a smarter and more ballanced split on the verses, when a scripture is added with "Split Long Verses"  selected.

Before:
image

After:
image


vassbo:
Would this work with more than two splits? E.g. if the verse should be splitted into 5 parts?

yurividal:
yes, it will.
Look how gracefully it splits Ester 8:9, which is the bible's longest verse

image

PR #2351, opened 2025-11-09
Optimize auto size

yurividal:
This attempts to optimize the delay caused by the auto-size in text boxes.
Currently, there is a signifficant noticeable delay in slide transitions when any text box in the template has any auto-size enabled.

This tries to solve it by implementing a cache, and optimizing the calculation of the font size. 

Autosize.ts now uses a single scale calculation instead of the previous binary-search loop. It measures once at the default size, scales according to the container, and performs light one-pixel adjustments.

It also drops the 500ms wait time in SlideItemTransition, and now we only pause (~120 ms) when the incoming slide still needs an auto-size pass (no stored autoFontSize), so cached slides appear instantly.

Updated slideHasAutoSizeItem to optionally check for missing autoFontSize, letting transitions know when the delay is actually required.

In TextBox, we  start auto-size on the next animation frame, ignore pure autoFontSize mutations when tracking changes, reuse any stored autoFontSize immediately, and cancel the RAF on destroy (this removes unnecessary extra autosize runs and keeps text sized correctly before the new calculation finishes)

This passes all my use-case tests, but more testing is probably needed when it comes to other elements such as fancy backgrounds and transition animations. (I don't use any of those).

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -74,15 +74,14 @@
     // timer updater
     let loaded = false
     let dateInterval: NodeJS.Timeout | null = null
-    let rafId: number | null = null
     onMount(() => {
-        rafId = requestAnimationFrame(() => (loaded = true))
+        loaded = true // ensure autosize runs before the first frame shows large text
+        if (itemElem) calculateAutosize() // pre-measure immediately so transitions do not flash oversized text
     })
     onDestroy(() => {
         if (dateInterval) clearInterval(dateInterval)
         if (loopStop) clearTimeout(loopStop)
         if (paddingCorrTimeout) clearTimeout(paddingCorrTimeout)
-        if (rafId !== null) cancelAnimationFrame(rafId)
     })
 
     // $: if (item.type === "timer") ref.id = item.timer!.id!


gladsonsam:
Looks like each branch is based on another PR branch instead of dev, which could make merging difficult.

yurividal:
> Looks like each branch is based on another PR branch instead of dev, which could make merging difficult.

Yeah, I only noticed that after I finished pushing the last commit.
The thing is, they are all sort of related, and I the entire reason I worked on these was to fix issues I was having with delays and with scriptures not parsing properly.
So, merging the latest branch, if they all get approved by @vassbo should be enough to implement all the fixes.

vassbo:
This is a good step @yurividal.
But the problem, and the reason for the 500ms delay is for when you don't have the `autoFontSize` pre-calculated, in the cases you have a custom output with a style template using auto size. In that case we have to calculate the auto size when outputting the text and can't do it before.

vassbo:
I'm sure it's generally better now, but you can see the text changing in the outputs with a custom style template. We need to finish the auto size, before changing the outputted text.

yurividal:
The key might be on the "slideHasAutoSizeItem" i tried to make it so that when a delay is actually needed, it would wait for the auto-size. but you are saying its not working on custom style for specific outputs. maybe i'm missing an edge case?

vassbo:
Create a new template with auto size enabled, and set that in your output style to "Override slide with template". Then make sure the show slides are different to that template, you will now see the text jump around when changing slides.

If you don't see it try increasing the transition duration to Fade 1 second.

vassbo:
Did you manage to recreate the auto size text issue?

yurividal:
no, not yet, let me try it now

yurividal:
I cant recreate it, but i think it might be because of this:
image


Looks like, in my case, this overwride takes precedence over the "Overwrite shows with template" 
Which is probably something we need to tackle in the  future... which overwrides have higher priority, and make it clear to the user.

Let me try without these group overwrides...

yurividal:
https://github.com/user-attachments/assets/b20335d8-61d9-4228-b55b-651895d9918c

no... still can't recreate... i created a new template, with orage font, and small text box, with auto-size shrink to fit, then enabled it as overwrite show with template, and removed all the groups from my song.

I can't get it to jump arround.

Even with transition fade 1s

https://github.com/user-attachments/assets/52ba3953-dc5f-4a06-870b-fcb79ef9cbad

vassbo:
Here is my example with auto size in the output style template:
image
image

It's most noticable in the output window, (shown in the bottom right):
https://github.com/user-attachments/assets/634f44dc-76b3-48ac-bb88-baa01fbd1dff

yurividal:
@vassbo i've watched the video, and also tried recreate the same exact template. I'm not sure what you mean by jumping arround. Do you say you can see the text in its original size, then see it change size? i can't see that in the video. =/

vassbo:
You can see the text is super large for a split second every time I change a slide, while it's resizing to fit before it settles into the actual size. Here it is at 0.1 speed, making it easier to spot, sometimes it's more noticable than other times:

https://github.com/user-attachments/assets/e5a4df93-ef7b-4f28-b678-3fc2ce128a77

yurividal:
> You can see the text is super large for a split second every time I change a slide, while it's resizing to fit before it settles into the actual size. Here it is at 0.1 speed, making it easier to spot, sometimes it's more noticable than other times:
> 
>  ezgif-6f17dbff18318437.mp4

Ohh, i see it now, on the very first slide in your video. I'll try to look into this tonight. maybe there is a way to condition a wait only if the slide needs it, and then, only wait as much as needed.

yurividal:
I think it mighth be related to the fact that you are using grow to fit. all my tests, i always used shrink to fit. i'll try to recreate it and then see what can be done to treat this case

yurividal:
@vassbo  can you check this latest commit? 

auto-sizing now runs as soon as the textbox mounts, preventing the first frame of a transition from showing oversized text.

also, removed the unused animation-frame guard so the initial sizing call happens immediately without waiting a frame.

vassbo:
Still the same.

yurividal:
Lol, I think the reason why im not seeing it is that my monitor runs at 30-60hz.

I'll try to borrow a 120hz monitor to try to recreate it here.

vassbo:
I have a 60hz one, so not sure why.

yurividal:
Ok, just spoke to a friend who is used to dealing with js
He agreed to try to help me sort this out. We are probably going to look into this in the next few days.

I'll probably close this pr, and stash my implementation, and start from your current dev branch again.

vassbo:
Okay, cool. I will probably create a new release today, so you can wait to clone the dev branch until after that.

PR #2358, opened 2025-11-11
split-reference edge-case fix

yurividal:
This fixes an edge case. 
Currently, If i select, for example, Deuteronomy 2:1-10, it correctly displays, splits long verses , and populates the slides and the refference shows Deuteronomy 2:1-10

But, if i select the entire chapter, the slides are split properly, but the reference shows Deuteronomy 2:1a-37b

This commit fixes this.


vassbo:
If the "Show letter suffix" is enabled, it won't show the letter in the reference. If I select Genesis 1:1b it will show Genesis 1:1.

yurividal:
yes, but if you select the ENTIRE chapter, it shows the letters... Trey Deuteronomy 2

This fixes it

yurividal:
Oh, i see what you mean. So, you want the reference to show the letters if Show letter suffix is selected?

vassbo:
Yeah, previously it showed Genesis 1:1b, but it no longer does with this change.

vassbo:
> Oh, i see what you mean. So, you want the reference to show the letters if Show letter suffix is selected?

Yes, I guess that makes sense, if you don't have the verse number enabled, but only show parts of a verse.

yurividal:
ok, i can look into that, for sure. ill update this and push again

yurividal:
in my opinion, the reference should never contain the suffix, only the verse itself. but up to you

vassbo:
I guess that makes sense.

PR #2360, opened 2025-11-11
More remoteshow improvements

gladsonsam:
Adds transpose to remoteshow


vassbo:
You can't import from the server in the frontend.

gladsonsam:
My bad, I guess we can have both transpose utils then?

vassbo:
Yeah, that's fine. It doesn't work in production that way.

PR #2378, opened 2025-11-12
expose logSongUsage dynamic value

yurividal:
This is so that companion can get variable and feedback about the status of logging of song usage


vassbo:
We should have an API only value for this, and not a dynamic value as it will never be used in-app.

Also if you run `npm install` you will get the latest packages, including the latest pdf worker. 😅

yurividal:
Cool, sorry for this.

It doesn't seem I can implement variables and feedback in companion if the value is an api-only.
At least, I couldn't figure out how to get it to work

vassbo:
There are two "custom" variables in Companion you can see as an example, `active_layers` & `active_styles`.

yurividal:
@vassbo  i'm really strugling to get the feedback and variable for this working on companion. I tried replicating exactly what you did with active_layers, but i just can't obtain the value.

vassbo:
Hmm..., there must be something missing then, I can give it a shot next week if you would like.

yurividal:
sure mate, if you don't mind. and thank you!

vassbo:
Didn't have time this week, but I'll try again next week. 😅

vassbo:
Here is the variable/feedback logic done: https://github.com/bitfocus/companion-module-churchapps-freeshow/commit/07a96316de532df4ba7a1b7ea059a321277d8c07
And in the app: https://github.com/ChurchApps/FreeShow/commit/51763b4ce2ecac1d0827469571b23bc91ddcf904

PR #2385, opened 2025-11-13
RemoteShow UI Improvements

gladsonsam:
This is generally what I was aiming for. It still needs some refinements, but let me know where it can be improved.

### DIFF
--- a/src/server/remote/components/tablet/TabletMode.svelte
+++ b/src/server/remote/components/tablet/TabletMode.svelte
@@ -29,6 +29,7 @@
     $: _set("layout", layout)
 
     $: totalSlides = layout ? layout.length : 0
+    $: outShowName = $outShow?.name || translate("remote.no_show_selected", $dictionary) || "—"
 
     ///////
 
@@ -220,6 +221,23 @@
         _set("outShow", $activeShow)
         // _set("outSlide", index) // ??
         // send("API:get_cleared") // ??
+    }
+
+    function openOutShow() {
+        const showId = $outShow?.id
+        if (!showId) return
+
+        send("SHOW", showId)
+        _set("active", { id: showId, type: "show" })
+        _set("activeTab", "show")
+        _set("activeShow", $outShow)
+    }
+
+    function handleLabelKeydown(event: KeyboardEvent) {
+        if (event.key === "Enter" || event.key === " ") {
+            event.preventDefault()
+            openOutShow()
+        }
     }
 
     // RESIZERS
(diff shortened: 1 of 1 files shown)


gladsonsam:
Going to revert SlideControls and keep how it was before, as it's much more usable.

gladsonsam:
I was thinking of adding an expand feature to the tablet mode. I haven't planned it out too much, but what I've noticed is that the show area is often wasted during the sermon time, where I am just using it for scripture. Instead it would be nice to use that area for a full scripture grid so I don't need to go between depths which does slow me down on tablet.

vassbo:
Just checked it out, it looks very clean!
Just one thing I noticed is the show buttons in the project looks too large on a computer, it might be good size for a smaller screen though.

vassbo:
We can merge once you think it's ready.

gladsonsam:
> Just one thing I noticed is the show buttons in the project looks too large on a computer, it might be good size for a smaller screen though.

Yeah, I've changed it back so they are the same size. 

I have a few more things like adding floating inputs, so it might be a few more days.

Also let me know what you think about the scripture overflow idea:

My main problem is I navigate too quickly on scripture and it zooms in, so I guess we could disable touch based zoom. Otherwise the scripture grid taking up show space would be nice.

vassbo:
Okay, nice. As for scripture you can resize the left panel to take up most of the space, maybe we could make also make it auto-expand? I think that would be the easiest, but if you have a good solution for tablets you can give it a shot.

vassbo:
I'm gonna create a beta release, so I think it's best to just merge what we have for now.

vassbo:
@gladsonsam The UI is so much cleaner now, thanks a lot. I did notice when going to the "Projects" page there are nothing showing up.

gladsonsam:
Oh, that shouldn't be happening. It worked for me at least...

I'll take a look later today

vassbo:
Okay, thanks. I just get this:
image

If it still works for you let me know, and I can check it out afterwards.

vassbo:
The active project works though so it's not a big deal.

gladsonsam:
Nevermind, it doesn't load but I should be able to fix it, will submit a PR soon

gladsonsam:
Strange, I can't replicate the issue on my dev machine, but on a VM, I did notice that the 'Example' project doesn't appear.

Do you have other projects? 

Here is what it is meant to look like:
image

gladsonsam:
Nevermind, I think its with the reciever. Will submit another PR

PR #2401, opened 2025-11-16
Fix output lyric overrides

yurividal:
today i found a weird behaviour.
On the preview main panel, the overwrite was working. i saw the word in italics. In my output preview (on the top right-side of the screen), i also saw the overwrite working. But, on the actual output, the word wasn't italic, the overwrite wasn't working at all.
Closing and reopening freeshow made it start working

this commig fixes output lyric overrides by deriving the show template directly from $showsCache.

This change replaces a non-reactive _show lookup inside Textbox.svelte with a reactive computation based on $showsCache, falling back to the focused or active show when needed. As a result, template style overrides now propagate to output windows immediately.

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -152,13 +152,43 @@
         // pick up template supplied by group overrides (if present)
         return $groups[groupId]?.template || ""
     })()
+    // track whether this textbox belongs to the first slide for the active layout
+    let isFirstLayoutSlide = false
+    $: isFirstLayoutSlide = (() => {
+        if (!ref?.showId) return false
+        const slideId = ref.slideId || ref.id
+        if (!slideId) return false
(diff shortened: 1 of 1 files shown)


yurividal:
The second commit makes sure that a style overwrite only gets applied to slides that follow the respective template. If use sets "Custom template on first slide" the first slide should inherit only the overwrites from its own template

vassbo:
Thanks, but the thing is, changes to slides/templates shouldn't update the output immediately, they should first update on refreshing the output. Also show/slide/group templates should update the show slides directly and not the output, the style templates should update the output.

yurividal:
Gotcha. I believe the issue is exactly on the refreshing the output part. Don't think it's working as intended. As far as I can see, the output preview should always reflect the actual output, but this wasn't being the case all the time

vassbo:
Okay, I will give it a test.

vassbo:
It looks to be working as expected, thanks.

PR #2403, opened 2025-11-17
auto-size caching and optimizing

yurividal:
Hi @vassbo 
As promissed, this is my second attempt to optimize the auto-size, to remove the visible delay on slide change when auto-size is enabled on any text box. 

Here is a brief summary of what we did:

We speeded up textbox autosizing by building it in stages. First we let each textbox hide itself until autosize finished, and we cached the measured font size so revisiting the same slide wouldn’t reflow everything. Next we pre-rendered hidden textboxes ahead of time so the cache was warm before the slide went live, and we stripped the old transition delay that was masking the work. We also had to fix style overrides by clearing any stale font sizes on template items and sharing the autosize cache across every output, so the secondary slide feed reuses the same measurements as the main screen.

I did a bunch of testing, with slow motion videos, and I was unable to observe the issue you were seeing last week, where the font was briefly appearing super big on the screen.

Let me know what you think of this new approach. Seems very quick on all my tests.

### DIFF
--- a/src/frontend/components/helpers/output.ts
+++ b/src/frontend/components/helpers/output.ts
@@ -780,6 +780,13 @@ export function mergeWithTemplate(slideItems: Item[], templateItems: Item[], add
     // should be reversed, but people have to invert the order of their template items order.
     templateItems = clone(templateItems) // .reverse()
 
+    if (resetAutoSize) {
+        templateItems.forEach((item) => {
+            if (!item) return
+            delete item.autoFontSize
+        })
+    }
+
     const sorted = sortItemsByType(templateItems)
     const sortedTemplateItems = clone(sorted)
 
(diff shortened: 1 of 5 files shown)


yurividal:
To do: 
I am still seeing a very noticeable delay on slide changes on my secondary output. (my OBS screen). 
From all my tests, it doesn't seem to be related to auto-size at all, nor is it related to NDI. Even If i have a lower-third template without auto-size, and disable NDI, the main screen still changes a LOT faster than the secondary output. This is even noticeable on the output previews, so its definetly happening inside freeshow.

I'll try to investigate further what could be causing this delay, but i think its unrelated to this PR.

vassbo:
Thank you for putting in the work! This looks good.

PR #2425, opened 2025-11-20
Normalise Payload

gladsonsam:
This should work and show all the Projects, but let me know if it doesn't.

### DIFF
--- a/src/server/remote/util/receiver.ts
+++ b/src/server/remote/util/receiver.ts
@@ -99,14 +99,17 @@ export const receiver = {
     PROJECTS: (data: any) => {
         if (!_get("isConnected")) return
 
-        _set("projects", data)
-        // newest first
-        _set(
-            "projects",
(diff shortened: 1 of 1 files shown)


vassbo:
Nope, no difference. Problem is that the projects/folders are not set to the root path "/" when parent folder is deleted, so if you place content in a folder and delete the folder it will not show up. I just apparently had all of my projects in deleted parent folders.

gladsonsam:
Might have to let you fix it. I can try take a look next week, but I can't replicate it yet and probably not before the next beta.

vassbo:
I can fix it. The folders must be deleted by another method than from the context menu. I guess it's not that common.

vassbo:
This is the fix if you are curious: https://github.com/ChurchApps/FreeShow/commit/8afb5015af5513c36ec2f7193d4145ca2b0add42

PR #2427, opened 2025-11-20
Feature -> sortable show headers

yurividal:
I have been working on this for a while, but had kinda left it, and finally decided to go back to it, since i had time today....

This implements an optional feature that can be turned on in the settings, which shows a sort bar above the show list.

The bar allows you to sort a list of shows by name, number, or last modified date.

under Settings -> General -> Enable Show Sort bar

Now, under shows, you see this:

https://github.com/user-attachments/assets/332198e4-522d-41cc-8636-547c0357502e


vassbo:
Fine, I guess it makes sense to have it. 😅
Thanks! It probably doesn't need a setting as people won't toggle it on/off.

But I updated the UI:
image

yurividal:
> Fine, I guess it makes sense to have it. 😅 Thanks! It probably doesn't need a setting as people won't toggle it on/off.
> 

Haha thanks. I remember many months ago, you said you didn't really see the usefulness of this feature, so that's why I implemented the option to disable it. Lol

vassbo:
Yep, the reason was because we already had sort from the context menu, and it took up space in the interface. But this is quicker, and now that I also added titles for different sections like "Categories" etc. it fits in better with the new UI.

PR #2435, opened 2025-11-22
Added search autoscroll and fixed CSS

gladsonsam:


gladsonsam:
Scriptures should be organised now. 

There is also a bug with the project folders where they don't expand all the time. It's hard to replicate so I might need some help.

vassbo:
So the projects inside closed folders did not get added to the tree, so they did not show up when the folder was opened.
I got that fixed and another minor issue.

gladsonsam:
Thanks, pretty happy with remoteshow now. 

I'll experiment with making scriptute easier but disabling zoom works pretty well for now.

gladsonsam:
You might be aware already, but the dev branch is now showing some new shows:
image

vassbo:
Interesting.. It's not there for me. That's the settings keys. Did you have any shows in the list previously?

vassbo:
Oh, I see them now.

vassbo:
Didn't sort correctly. Thanks for letting me know, I'll fix that. Maybe it is like that in the beta as well?

gladsonsam:
Yeah, but I only saw it now, that's interesting.

vassbo:
I'm glad you spotted this. Thank you!

It happened when restoring the settings.json file, I passed that to `restoreShows` instead of `restoreStore`. 😅

gladsonsam:
That explains why I only saw it now. Glad I could help. Let me know if this is worth pursuing:

In tablet mode, the header in scripture has an expand button which puts it to this view:
image

This can then be compacted by pressing the button or by navigating to another tab. I still have lots more to do, so it's cohesive with the main app UI. I think it could work, and it doesn't add that much complexity to the RemoteShow for people who don't want it.

vassbo:
Okay, sounds like it can work. :)

gladsonsam:
I thought about it for a bit and couldn't figure out a good UX. So I spent a bit of time creating a better TabletMode from the ground up:

IMG_1410

I know it would be quite a change, but I think it's good to have a good amount of control in remoteshow, especially for devices with the real estate. Obviously, it wouldn't do everything or even have all the drawer tabs, but the basics like shows, scripture and functions. Tasks such as adding an item to a project is easier this way and I just think its better to be cohesive with the main app frontend.

It's up to you if we include it in the app, but I just thought I would try a draft first.

vassbo:
Wow! That is better, yes, the main reason I had not done that was because it would require a lot of work, and likely not be used that much. But this looks good.

If I were to start again from scratch, I would make the drawer use the API, so it could be detached into it's own window and also used without changes in the remote like this. But now the drawer code has to be written twice, which works I guess.

Issue #2415, opened 2025-11-19
Scripture is broken by text formatting


PavelS-CZ:
**Describe the bug**
Scripture is broken by the default formatting algorithm of Freeshow.
Some parts of scripture are greyed out and formatted differently because of symbols included in the bible text itself.

**Reproducing**
Insert the scripture text, and freeshow automatically reformats it, which breaks the text

**Version (Optional)**
 - OS: Windows 11
 - FreeShow Version: 1.5.3-beta2

VIDEO: https://drive.google.com/file/d/1s9xPCR67AnIMr-wa94BVLp6LC9up9oIf/view?usp=sharing
BIBLE FILE: [CSP.xml](https://github.com/user-attachments/files/23635146/CSP.xml)
_

> <v n="19">
> Neboť Boží Syn, Ježíš Kristus, který byl mezi vámi skrze nás vyhlášen – mnou, Silvanem a Timoteem – nebyl ‚ano‘ i ‚ne‘, nýbrž v něm se uskutečnilo ‚ano‘.
> </v>
> <v n="20">
> Neboť všechna Boží zaslíbení, kolik jich [jen je,] v něm [jsou] ‚ano‘. Proto [je] také skrze něho [řečeno] ‚amen‘ k slávě Bohu skrze nás.

_

TEXT: https://biblenow.net/cs/bible/cesky-studijni-preklad/novy-zakon/2-korintskym/1

yurividal:
Did this always happen, or is it only with 1.5.3? Does it work on other older versions?

yurividal:
Ok, i found the problem. Its in the split-verse logic. I'll try to fix it now. 

Also, should the words in between swuare brackets be grey and italic by default?
My oppinion is no, since you can add a custom regex to make this grey and italic if you want, but i honestly don't know the Czech language and don't know if this is common.

vassbo:
Thanks @yurividal.

I think it can be gray by default, but the user should be allowed to do any custom styling for it with template, as you just implemented that feature.

PavelS-CZ:
Thank you a lot for fixing this @yurividal 
That bug kind of scared me when I saw it, since this translation is often used for preaching. 

Yeah, it would be great if there was a toggle for: 
 • the text being grayed out, as the translation shouldnt be
 • italic 


(And quick fun fact, those squared brackets are used only by Bible translators to indicate which words were not included in the original manuscripts but had to be added for proper readability)

yurividal:
[...]

PR #2447, opened 2025-11-25
Tighten scripture template overrides

yurividal:
This attempts to fix https://github.com/ChurchApps/FreeShow/issues/2415#issuecomment-3577652806

Here is what this PR implements:

- Store the scripture template id on each generated show so Textbox always knows which overrides belong to that scripture block.

- Update the template resolution order to prefer output-level scripture layouts before falling back to the stored show template

There is also some code lynting (indentation) auto-applied by my vscode. (sorry)


vassbo:
Thank you!

Issue #2169, opened 2025-09-23
Rework Transitions


vassbo:
**Describe the feature**
Better/smoother text transitions.
Also the current method sometimes breaks in rare occasions: https://github.com/ChurchApps/FreeShow/issues/1006

**Additional context**
Suggestions here (pretty similar):
- Scrolling Slide Text: https://github.com/ChurchApps/FreeShow/issues/1695
- Fading Text Slide: https://github.com/ChurchApps/FreeShow/issues/1767
- Scrolling Text Style: https://github.com/ChurchApps/FreeShow/issues/1876

prowhiz:
I'm not sure if this is already part of the original suggestion, but I noticed that the transition usually includes the background especially the fade transition. I'd prefer a state of affairs where the background remains constant and only the text transitions.

jerm00:
> I'd prefer a state of affairs where the background remains constant and only the text transitions.

I'm not sure what you mean, this has always been the case unless you add a different background on a slide and then click it, or manually change the background. The backgrounds can have a different set of transition effects if desired, a feature added maybe a year ago.

jerm00:
@vassbo i have an idea that you could use as a workaround for implementing a proper crossfade (assuming it might not be possible to implement it normally):

if someone selects crossfade as their text transition, what if you made it so a comparison is done between the current slide and the next, and if an identical object appears on both slides, then don't fade it from the first one, or fade only the parts that change. for example, if a textbox has different text on both slides, but the same background color/opacity and a blur effect, fade the text as it is currently implemented, but leave the empty textbox with the background color/filter so that a fading/adding back isn't visible. i think that would appear to give of the appearance of crossfading.

vassbo:
@prowhiz Sounds like you added the media as slide elements and not as background then.
@jerm00 That doesn't work with the current system. Also using proper backgrounds/overlays overcomes that problem.

taenologue:
How about trying WebGL-based transitions using PixiJS or Three.js?
They could provide smoother, GPU-accelerated text and slide animations.

[PixiJS : https://github.com/pixijs/pixijs](https://github.com/pixijs/pixijs)
[Three.js : https://github.com/mrdoob/three.js](https://github.com/mrdoob/three.js)

vassbo:
WebGL is not bad for advanced text renderings, but might be overkill for simple transitions.

frederickjh:
[...]

PR #2451, opened 2025-11-26
Prevent scripture metadata from flashing during slide blanks

yurividal:
This solves a minor issue i noticed this weekend. 

Since i don't use any transitions, i noticed that my bible metadata and referrences were brifely flashing on the screen at each slide change. Since this data is static, there is no need for us to blank and re-draw them, so i wanted to implement a way to detect if a text is unchanged between 2 slides, and just keep it rendered, instead of redrawing it.

In my case, the scripture reference and metadata textboxes were tearing down during every slide blank because they lacked stable IDs, so even unchanged content disappeared momentarily. 

This patch keeps "carry-over clones" mounted through the blank and releases them only after the new slide is visible, which stabilizes the static UI while still blanking changing verse text.

Here's what was done:

- Added a persistent hand-off path inside SlideContent.svelte so unchanged textboxes can bridge the “blank” interval: when both the outgoing and incoming slide have content, we now take a snapshot of the previous render (`lastRenderedItems`) and run a two-step match. First we compare by ID via `buildSignatureMap`, and when IDs are missing (typical for scripture metadata), we fall back to hashing the textbox payload (`getItemSignature`) and decrementing `nextSignatureCounts` so each identical textbox pairs once.

- Any matched items get cloned into `carryOverItems` along with the exact `outSlide`, `slideData`, `currentSlide`, `lines`, and `currentStyle` state that produced them. Instead of dumping these clones once the blank starts, we gate their visibility behind a new `persistentHold` flag. The template now renders carry-overs whenever either the blank is active or that hold is still on.

- During a blank: we hide the main `currentItems`, leave the carry-overs onscreen, switch the backing slide data by calling `applyCurrentSlideState()`, and only tear down the carry-overs via `scheduleCarryOverRelease()` after the new slide has begun fading in (`waitToShow`). The release helper also enforces a minimum delay so the carry-overs survive long enough even when transitions are nearly instantaneous.

- When no blank is needed we call `clearCarryOverItems(true)` immediately. the helper also cancels any pending release timeouts to avoid later leaks.

### DIFF
--- a/src/frontend/components/output/layers/SlideContent.svelte
+++ b/src/frontend/components/output/layers/SlideContent.svelte
@@ -146,15 +146,24 @@
                 currentStyle: clone(currentStyle)
             }
             lastRenderedItems = []
+            clearCarryOverItems(true)
+            if (show) {
+                show = false
+                if (import.meta.env?.DEV) console.debug("[SlideContent] forced hide due to empty slide")
+            }
             return
         }
 
+        cancelPendingBlank()
         scheduleAutoSizePrecompute(currentSlide.items)
 
         const previousItemsSnapshot = lastRenderedItems.length ? lastRenderedItems : currentItems
         const previousSignatures = buildSignatureMap(previousItemsSnapshot)
         const nextSignatures = buildSignatureMap(currentSlide.items)
-        const nextSignatureCounts = buildSignatureCounts(currentSlide.items)
+        const nextItemsById = buildItemIdMap(currentSlide.items)
+        let fallbackTransition: any = null
+        const previousCount = previousItemsSnapshot.length
+        const nextCount = currentSlide.items.length
 
         // get any items with no transition between the two slides
         let oldItemTransition = currentItems.find((a) => a.actions?.transition)?.actions?.transition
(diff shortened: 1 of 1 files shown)


yurividal:
I have done all the tests with my current setup, including multiple diferrent templates, and also tested with songs, to make sure this doesn't break anything. But, please test this with diferent animations, and backgrounds, just to make sure we don't break anything. (It shouldn't but just in case...)

vassbo:
Thanks.

When "Fade" is enabled the transition does not happen sometimes when going from an empty slide to a slide with text.

Also sometimes if you have two slides with identical text, and a transition enabled between you do want the transition to be visible. Only if the transition is set to "None" like in your case do you wanna not re-render.

yurividal:
ok, let me try to adjust to that.

yurividal:
I haven't been able to implement this properly, without causing the referrence and metadata to blink between slides. i'll close this PR for now, and try this again another day

vassbo:
Okay, thanks for giving it a shot. I do want to rework transitions someday which should fix this also (or atleast make it easier to maintain): https://github.com/ChurchApps/FreeShow/issues/2169

Issue #2454, opened 2025-11-26
Some API bibles have hardcoded <br> which break split verse logic


yurividal:
Some API bibles, such as King James, have hardcoded html tags, such as `<br>` which cause the verse split to incorecrtly display the verses on the verse picker.:

````
[...[
{"value":"<br><br> And I spake unto you at that time, saying, I am not able to bear you myself alone:</br></br> ",
[...]
````

<img width="1089" height="185" alt="Image" src="https://github.com/user-attachments/assets/d83c82c4-0e90-4ada-b880-c357c177a742" />

Opening this bug report for tracking purposes. 

I'll try to implement a fix for this

PR #2455, opened 2025-11-26
Sanitize line break tags on API bible verses

yurividal:
This fixes https://github.com/ChurchApps/FreeShow/issues/2454


vassbo:
Thank you! It's appreciated.

PR #2462, opened 2025-11-27
Fix slide flicker

yurividal:
New attempt to fix the case where content that doesn't change is blinking on slide transitions, when no transition animations are set.

This should consider when items HAVE transitions, and if they do, it will follow the old procedure. 

@vassbo  is this better?


vassbo:
This looks very good, thanks again!
Seems to not break anything, at least in my testing.

vassbo:
I've had more people report slide flicker issues now than previously for some reason.

vassbo:
If one of the items has a transition it will affect all the other items with no transition.

PR #2463, opened 2025-11-27
Add visual distinction for sub-items in panels

yurividal:
Small UI/UX improvement, that makes sub-items distinct from parent items in drop-down menus. this makes it clear what options are sub-options.

Before:
image

After:

image


vassbo:
That's a good idea. thanks. I did end up implementing it in the InputRow component instead so it's the same everywhere the "arrow" is used. (Profiles/styles/timers/effects etc.)
https://github.com/ChurchApps/FreeShow/commit/f84c3ad45efd72aab5c22fa7a9f91d307bc34fb3

And tweaked the UI:
image

yurividal:
It's hard to tell from the picture, I'll have to test to check, but doesn't keeping the same color still make it a bit hard to tell? Or do you think the slight padding to the right is enough?

vassbo:
I think it's much better now at least. I didn't like the low contrast with the grey color, and having the same background makes all inputs look more consistent.

vassbo:
One alternative is to indent it even further.

PR #2480, opened 2025-12-01
Show all Bible versions in collections

BogdanIvchenko:
### Summary
When using Bible collections, this update displays all translation versions for each verse simultaneously in the verse selection UI, instead of showing only one version at a time.

### Changes
- **Main app**: Modified verse display in collections to show all versions stacked per verse
- **Remote viewer**: Same multi-version display for the web remote (localhost:5510)
- **Visual distinction**: Each translation has a unique color-coded left border using dynamically generated colors that match FreeShow's theme
- **Compact layout**: Removed padding and translation labels for a cleaner look while keeping verse numbers

### Why
When comparing translations or preparing slides with multiple Bible versions, users can now see all versions at a glance without switching between them. This makes it easier to select the right verse when following the preachers who use different languages. 

### Screenshots
Screenshot 2025-11-30 at 8 05 10 PM
Screenshot 2025-11-30 at 8 58 22 PM
FreeShow Remote

### Testing
- Tested with 3-version and 2-version Bible collections
- Single Bible selection still works as before
- Grid mode unaffected (feature only applies to list mode)
- Remote viewer displays correctly

### DIFF
--- a/src/frontend/components/drawer/bible/Scripture.svelte
+++ b/src/frontend/components/drawer/bible/Scripture.svelte
@@ -39,6 +39,107 @@
     // timeout is to load drawer tab "instantly", before loading scripture
     $: setTimeout(() => loadScripture(previewBibleId), 10)
 
+    // Load all scriptures in collection for multi-version display
+    $: if (isCollection) {
+        activeScriptures.forEach(id => {
+            if (id && !data[id]) {
+                loadScriptureForCollection(id)
+            }
+        })
+    }
+
+    async function loadScriptureForCollection(id: string) {
+        if (!id || data[id]) return
+        try {
+            const jsonBible = await loadJsonBible(id)
+            data[id] = { bibleData: jsonBible }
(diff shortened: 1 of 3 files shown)


vassbo:
Looks good! Thank you.

gladsonsam:
This looks great! I wonder if we could make the refresh button toggle through the translations, including an “All” option, so we can still view just one if needed.

vassbo:
@gladsonsam That's not a bad idea.

vassbo:
@BogdanIvchenko Sorry, I realized this didn't work with the "Split long verses" option, I commented it out temporarily. Would you mind checking that out?

vassbo:
Also we should have a toggle to turn it off, or something like @gladsonsam suggested, as it's quite hard to navigate when you don't need to see all the variances (like I guess is the case most of the time).

BogdanIvchenko:
I made the PR#2517 to address this.

Issue #2486, opened 2025-12-01
Request: NDI Output name and group


eeickmeyer:
Some networks with NDI require a specified output name and group in order to prevent NDI from flooding the network, and to keep NDI isolated from other items on the network. For instance, in OBS with the DistroAV plugin installed:

<img width="655" height="778" alt="Image" src="https://github.com/user-attachments/assets/cd309c7f-70f7-4b8a-a988-c57b003a9627" />

As you can see, this would be super beneficial so that networks that require this information can be permitted.

Right now, the only workaround is to route it into OBS via the DistroAV NDI plugin which can then route via the name or group required by the NDI network.

vassbo:
That seems like it would be helpful. I think we should probably add a groups option into the inputs discovery as well.

vassbo:
Let me know if it works or not: https://github.com/ChurchApps/FreeShow/releases/tag/v1.5.5-beta.1

eeickmeyer:
Unfortunately, it's hard to tell as neither field persist between restarts. Both fields require an application restart to take, per the dialog that pops up, but after restart neither is how I left it.

vassbo:
Did you save before closing, after changing the value?

vassbo:
In my testing it looks to be working as when I set a group to a NDI output, it only shows up in the inputs when I set it to that same group.

vassbo:
I made it change the visual state to not saved when changing the group as I'm sure you were not going to be the only one encountering that issue. ;)

eeickmeyer:
Probably a good idea. A lot of people are going to be coming from ProPresenter which saves states like that on-the-fly, so having no indication that they need to save is going to throw them.

PR #2491, opened 2025-12-02
NDI Groups

vassbo:


vassbo:
https://github.com/ChurchApps/FreeShow/issues/2486

PR #2492, opened 2025-12-02
Allways suffix on grid view

yurividal:
This makes it so that when in grid mode, we always display the letter suffix on the grid. 

Stil respects the "Show letter suffix" toggle on the slides itself, but on the verse picker, we should always show the suffix)


vassbo:
Thanks a lot!

PR #2498, opened 2025-12-04
Fix song number search to support alphanumeric values

yurividal:
I noticed that the search wasn't really working as expected, when alphanumeric song numbers:

Before this commit:
image

After this commit:
image

I don't think this breaks any current feature. at least not from my tests.


vassbo:
Looks good, thanks.

Issue #2501, opened 2025-12-05
Few more issues with new scripture template system


yurividal:
I've found a few more issues with the new template model:

1) I can't find the option to include reference on first slide. was this removed on purpose?
2) To work arround 1), i created a new template, with {scripture_refference_full} centered. And, spplied the option to "Use custom template on first slide", but, now, i get 2 title slides. 1 with the bible_name and one with the reference.
The first one also doesn't pick up {scripture_name} 
<img width="1953" height="846" alt="Image" src="https://github.com/user-attachments/assets/0dff8ba5-d7c4-4b62-a725-55ce7fb1de47" />

3) The numbering on split-verses is all messed up. As you can see on the screenshot above, i have 2 number ones, and the third part of number 1 doesn't have the number. 

4) {meta_title} doesn't seem to behave as the new {scripture_*} tags. Its using the old model, right?

5) I couldn't find a way to change the size of {scripture_number} without changing the size and colour of {scripture_text} like you did in the default template.

<img width="2454" height="1184" alt="Image" src="https://github.com/user-attachments/assets/519b2171-7822-4089-9f43-e28075061b60" />

vassbo:
1. Use "Custom template on first slide" with dynamic values.
2. It needs to be applied before you convert to show.
3. I have not seen that myself, maybe it's something with your template.
4. It should behave the same in the drawer, and when converting to a show.
5. Any styling to {scripture_number} will be transferred to the number in the drawer.

vassbo:
Remember to test any changes to the template directly in the drawer, and then with convert to show. Not from a template to a show directly!

yurividal:
<img width="2185" height="1219" alt="Image" src="https://github.com/user-attachments/assets/a0a7de71-7d5d-46a1-8ad4-494c1104fcf4" />

I did all you said, but im still getting 2 title slides. I'm not sure where that first slide is coming from. I never had that there.

Also, i still can't find a intuitive way to change the size of the verse number.

https://github.com/user-attachments/assets/c279ca2d-65fd-4cd5-9dfc-81e29933e32f

vassbo:
Oh, I forgot to check for that actually, in my testing I had the option turned off. So set to any template without the {scripture_ values and then turn off the "Reference on first slide" option, and set back to the new template.

See how the {scripture_number} is done with the default template, it should work.
[...]

PR #2502, opened 2025-12-05
Preserve text selection when changing font size in template editor

yurividal:
This is my attempt to improve what was discussed [here](https://github.com/ChurchApps/FreeShow/issues/2501)

- Add capture phase mousedown handler to save selection before input focus
- Check if selection is inside .edit element before updating selection variable
- Prevents selection from being overwritten when clicking on style inputs
- Allows applying font size changes to selected text only


vassbo:
Doesn't this do the same as the already existing code? I don't get what's different.

PR #2503, opened 2025-12-05
Add auto-convert feature for scripture templates - POC

yurividal:
This is a proof of concept about what we spoke earlier. I added a button to the template editor, that attempts to auto-convert an old scripture template to the new scripture mode.
Because the previous mode was only based on order of elements, i have implemented 4 different strategies to try to "guess" what the old template was being used for. The button cycles through the 4 strategies, in the hope that one of them will match what the user wants - or close to it.
Here are the methods:

Combined (Smart) - Uses weighted scoring from position, size, and content analisis
Position-based - Items at top = verse, items at bottom = reference
Size-based - Largest text = verse, smallest = reference/version
Content-based - Analises text patterns (chapter:verse notation, Bible version names, etc.)

Before converting, we store the original items to enable retry with different strategies and enable undo.

If you like this strategy, we can then try to add the convert button into a location that makes sense

### DIFF
--- a/src/frontend/components/edit/scripts/convertScriptureTemplate.ts
+++ b/src/frontend/components/edit/scripts/convertScriptureTemplate.ts
@@ -1,0 +1,466 @@
+import type { Item, Template } from "../../../../types/Show"
+import { clone } from "../../helpers/array"
+import { getStyles } from "../../helpers/style"
+
+interface ItemAnalysis {
+  index: number
+  item: Item
+  type: "text" | "other"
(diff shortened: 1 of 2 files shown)


vassbo:
If were are gonna do the auto convertion we have all the data to do it correctly and not "guess", but that requires a lot of work, and it will be just temporary anyway, as the old system will be removed in a future update.

I gues most people does not use custom templates, and the people that do should be able to figure out how to customize them again with the new system.

PR #2507, opened 2025-12-06
Add balance text lines option to templates

yurividal:
This PR adds a button checkbox under Template -> Edit -> Slide -> Text

This checkbox enables balanced text CSS into the template. 

This allows for long sentences to be evenly balanced.

Before:
image

After

image


vassbo:
That's much better visually, I actually think we should have this on by default always and not through a setting.

yurividal:
> That's much better visually, I actually think we should have this on by default always and not through a setting.

Cool. I just didn't want to make It default in case it broke something for someone. Your call.
It's much better for me. I found myself fixing lines every single week, whenever we sang a new song that we hadn't sang in a while

yurividal:
@vassbo 

https://www.facebook.com/groups/freeshowapp/posts/25502448206108788/?comment_id=25504347045918904

Maybe this should indeed be made optional? Perhaps default to enabled. 

I think this user is right, it doesn't realy make sense to have this on lower thirds, with bible readings.

vassbo:
I have already fixed this last week actually.

PR #2537, opened 2025-12-11
More RemoteShow Improvements

gladsonsam:
- Better Auth UI
- Media controls that match the frontend, with online mostly working (still need to figure out thumbnails)
- Audio Tab with a mixer section

I will clean up the code and conduct further testing, and let you know when it is ready to merge.
Let me know if you have any suggestions. Made some API additions so hopefully didn't break anything

### DIFF
--- a/src/server/remote/components/pages/Scripture.svelte
+++ b/src/server/remote/components/pages/Scripture.svelte
@@ -299,135 +299,30 @@
         searchInput.select()
     }
 
-    function formatBookSearch(search: string): string {
-        return search
-            .toLowerCase()
-            .replace(/\s/g, "")
-            .replace(/\./g, "")
-            .normalize("NFD")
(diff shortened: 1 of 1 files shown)


vassbo:
As for things like the `scripture-utils.ts` I think it's better to use the existing frontend code, to send and receive a request. The abbreviations should not need to be hard coded on the server as the `json-bible` package handles that, (also you for example do `.normalize("NFD").replace(/\p{Diacritic}/gu, "")` instead of all the hardcoded `formatBookSearch` replacements, but this should also already be handled by the `json-bible` package.)

vassbo:
I know it's gonna be hard to maintain this in the future, but it seems nice, and I guess it's relatively useful.

gladsonsam:
Sorry ran out of time to work on this. I hope the changes were alright.

When I get time, I will submit a PR to add split long verses etc.

PR #2560, opened 2025-12-15
Fix show all updating bug

BogdanIvchenko:
Before, if you changed the chapter and book and had "show all" selected, it would only update one of the versions to that book and chapter. Now it updates all of them.


vassbo:
Thanks! :)

Issue #2561, opened 2025-12-15
Verses per slide and verse split are broken in 1.5.5 when spawning multiple chapters


yurividal:
I'll try to take a look at this. should be an easy fix.

Someone in my team noticed that the verses are getting messed up when the scripture involves multiple chapters. If user splits a verse, it messes up even more.

Example:

<img width="2004" height="877" alt="Image" src="https://github.com/user-attachments/assets/f05bd46a-6843-4c89-8904-a7b692388032" />

<img width="500" height="319" alt="Image" src="https://github.com/user-attachments/assets/1ae7d990-e5ae-48c7-b6e7-a0180570fd46" />

<img width="1007" height="343" alt="Image" src="https://github.com/user-attachments/assets/efa7c823-3729-41c2-b4c4-44ac083908f6" />

This only happens when the scripture spans multiple chapters, which is a recent addition to the code.

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.5.6-beta.1

PR #2562, opened 2025-12-15
Fix the verse division when text spawns multiple chapters

yurividal:
Fixes https://github.com/ChurchApps/FreeShow/issues/2561

The previous slide generation logic would flatten the list of selected verses, losing the chapter context. This caused verses to be grouped incorrectly (example: verse 1 from each chapter would be grouped together).

The splitContent function has been updated to process verses sequentially, ensuring that all verses from one chapter are displayed before moving to the next. This maintains the correct biblical order when generating slides from multi-chapter passages.


vassbo:
Thank you! :)

PR #2564, opened 2025-12-15
Implement folder delete option

yurividal:
Currently, when deleting a folder in the project list, all projects inside were automatically moved to an "Untagged" folder (parent). This could be frustrating for users who wanted to delete a folder along with all its contents in one action, rather than having to manually delete each project first.

This PR adds a confirmation popup when deleting a folder that contains projects or subfolders. Users now have two options:

image

Delete folder only - Keeps the existing behavior where projects are moved to the parent folder
Delete folder and projects - Permanently deletes the folder and all projects/subfolders inside it


vassbo:
I see. How about this?:
image

vassbo:
Or maybe better it can delete the contents always, but show a confirmation. Because you can move the content or rename the folder otherwies.

yurividal:
Yeah, up to you!
I just needed the ability to delete a folder and it's content. However you think is best to have it.

vassbo:
Cool, I think it makes most sense to always delete the content as well.

Issue #2585, opened 2025-12-20
Regarding multiple scripture verse selection


ole-oss:
**Describe the bug**

To instantly display verses, 
1. ctrl+f 
2. type the chapter and minimum 2 verses
3. press enter
4. ctrl + right/left to move to next/previous verses

A problem occurs when moving to next batch of verses by pressing ctrl+left arrow, where a single verse gets skipped every time. It also happens when i click on the next slide arrow in the drawer tab.

**Reproducing**

Refer to the video

**Screenshots (Optional)**

https://github.com/user-attachments/assets/592cafbe-de2b-4be7-818b-8c4b7ccefcc3

**Version (Optional)**
 - OS:  Win 11
 - FreeShow Version: 1.5.6 beta 1

**Additional context**

ole-oss:
After a little more experimenting, i found out that if you have n number of verses selected to display, then the number of verses skipped is n-1.

Also this does not happen if i convert to show and press ctrl+right once. The skipping of verse(s) starts to occur again from the second time.

vassbo:
It worked perfectly fine when it was not outputted apparently, which is why I didn't catch this.

vassbo:
It was apparently not sorting the verses in the correct order, causing this.

PR #2602, opened 2025-12-26
fix next slide indexing

rchallenger:
The `currentVerses` list wasn't always holding sorted values, but the algorithm assumed it was. So now the list is sorted initially. ToDo: look for other related sensitive places in code.

Reference for the issue discussion: https://livechurchsolutions.slack.com/archives/C074Q696S8L/p1766584600906569

### DIFF
--- a/src/frontend/components/drawer/bible/Scripture.svelte
+++ b/src/frontend/components/drawer/bible/Scripture.svelte
@@ -776,7 +776,7 @@
         if (!activeReference.book) return
 
         // Check if we're dealing with split verses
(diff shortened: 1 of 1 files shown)


vassbo:
Thanks, but this was already fixed a week ago on the dev branch: https://github.com/ChurchApps/FreeShow/commit/4fb44165e3bc2b32d59f9dbe54029f61bfb98fe1
Also this would not sort properly in all cases.

Original issue: https://github.com/ChurchApps/FreeShow/issues/2585

PR #2630, opened 2026-01-03
Add verse select

gladsonsam:
I have been using RemoteShow for a while but found it inconvenient when I couldn't display verses together. This adds a checkbox and works on both mobile and tablet. It can also be activated by long press to make it easier to use on touchscreens. Let me know if the UI can be improved. 

image

I don't really know where a good place is to put the present button on mobile.

Also edited the API to allow multiple verses.

### DIFF
--- a/src/server/remote/components/pages/Scripture.svelte
+++ b/src/server/remote/components/pages/Scripture.svelte
@@ -994,11 +994,6 @@
                             <Icon id={$scriptureMultiSelect ? "close" : "check"} white />
                         </Button>
                     {/if}
-                    {#if depth === 2 && $scriptureViewList && $scriptureMultiSelect && $selectedVerses.length > 0}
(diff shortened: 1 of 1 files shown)


vassbo:
Nice, I guess that's fine. Maybe only show the checkboxes if holding down to select a verse, if not present the clicked verse.

gladsonsam:
> Nice, I guess that's fine. Maybe only show the checkboxes if holding down to select a verse, if not present the clicked verse.

I think that's how it's implemented, unless I misunderstood. I have also moved the play button to the right.

vassbo:
Okay cool. Is it ready?

gladsonsam:
Should be good. It works fine in testing and would be nice to have in the next build. 

I do want verse highlighting to work for multiple verses but I will submit another PR for that if I can get it working.

PR #2648, opened 2026-01-08
fix: search all lines for customType markers in scripture auto-sizing

yurividal:
This PR fixes https://github.com/orgs/ChurchApps/discussions/728

The reason that it was happening is:

The customTypeRatio calculation was only searching the first line (lines[0]) for verse numbers marked with customType: 'disableTemplate'. In templates where verse numbers appear in other lines (e.g., line 1), this caused the ratio to default to 1.0 instead of the correct value (e.g., 0.6 for 60px verse numbers).

This resulted in inconsistent auto-sizing where some verses would be calculated with extremely small font sizes because the algorithm treated the verse number as if it should be the same size as the verse text.

Fixed by iterating through all lines to find customType markers in both:
- deriveCustomTypeRatio() function
- calculateAutosize() function

Now correctly calculates the verse number to text ratio regardless of which line contains the verse numbers.

Let me know what you think of this approach.

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -1,844 +1,684 @@
 <script lang="ts">
-  import { createEventDispatcher, onDestroy, onMount } from "svelte";
-  import { OUTPUT } from "../../../types/Channels";
-  import type { Styles } from "../../../types/Settings";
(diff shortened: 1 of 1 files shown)


vassbo:
Nice, this looks fine to me.

Issue #2617, opened 2025-12-29
Lower Thirds scripture template inconsistent


ehow-nacharam:
**Describe the bug**
The color of the text is different for multiple verses. If I have only one verse the text color is off-white and the visibility is very low. Also there is some issue with the auto size feature.

**Reproducing**
While presenting a scripture lower thirds using the FreeShow's default template

**Screenshots (Optional)**

<img width="1841" height="1031" alt="Image" src="https://github.com/user-attachments/assets/4b7df2df-ca17-4c2b-969f-84764eff68bb" />

<img width="1830" height="1029" alt="Image" src="https://github.com/user-attachments/assets/ac8ce5f7-cbf4-4e12-b625-93773d4dd4d3" />

**Version (Optional)**
 - OS: Windows
 - FreeShow Version: 1.5.5

thebiblelover7:
I am not being able to reproduce this... Are you using the actual default template, and have you modified it at all?

keeandev:
I'm running into the same issue, I turned off the border and made the scripture reference black, and removed the border but the second verse keeps the border. 
<img width="2972" height="636" alt="Image" src="https://github.com/user-attachments/assets/a1a8c3d6-e63f-4dc2-bbb9-542d73785839" />

<img width="2972" height="636" alt="Image" src="https://github.com/user-attachments/assets/1046f926-4e6b-4a8a-91d3-c9a48877d524" />

yurividal:
Are you guys using the new scripture template model? trying to reproduce this here

yurividal:
<img width="2829" height="479" alt="Image" src="https://github.com/user-attachments/assets/6ce5032e-e0cd-410a-ae23-195cb2726100" />

I'm unable to reproduce, using the latest dev branch code. 

This is how my template looks:

<img width="1993" height="352" alt="Image" src="https://github.com/user-attachments/assets/caf04fa8-e340-4e75-9804-243519aee248" />

Jehne07:
Hi, the issue is most likely that when you choose a template for Scripture in the Styles settings, it’s still being overwritten the old way. (You’re probably using it just for another output, for example, for live streaming.)

yurividal:
@Jehne07 THATS IT!
I was able to reproduce it now. Only happens when there is a second output, and that output is overwriting the template

<img width="2009" height="575" alt="Image" src="https://github.com/user-attachments/assets/5796bcf8-22c9-4ed0-8163-f9419ec8df9f" />

[...]

PR #2653, opened 2026-01-09
Fix scripture template override applying wrong styles to verses 2+

yurividal:
This fixes https://github.com/ChurchApps/FreeShow/issues/2617

When using 'Overwrite Scripture with Template' in output styles, verses beyond the first were getting matched to the wrong template text object (falling back to text[0] which is the verse number template instead of verse text template).

Now intelligently matches verse text objects to {scripture_text} placeholder and verse numbers to {scripture_number} placeholder based on customType flag.

### Problem
When using "Overwrite Scripture with Template" in output styles, scripture slides with multiple verses displayed incorrect formatting:
- Verse 1: Correct styling (black text)
- Verses 2+: Wrong styling (grey text from verse number template)

### Root Cause
In `output.ts`, when applying style templates to scripture slides, the template text matching logic fell back to `text[0]` for verses beyond the first. Since `text[0]` is the verse number template (grey, 40px), all subsequent verse text got styled incorrectly.

### Solution
Added intelligent template matching that:
1. First tries exact index match
2. If that fails on scripture templates, matches by placeholder type:
   - Verse numbers (with `disableTemplate` flag) → `{scripture_number}` template
   - Verse text (without flag) → `{scripture_text}` template
3. Falls back to `text[0]` only as last resort

### DIFF
--- a/src/frontend/components/helpers/output.ts
+++ b/src/frontend/components/helpers/output.ts
@@ -4,7 +4,7 @@ import { OUTPUT } from "../../../types/Channels"
 import { Main } from "../../../types/IPC/Main"
 import type { Output, Outputs } from "../../../types/Output"
 import type { Resolution, Styles } from "../../../types/Settings"
-import type { Item, Layout, LayoutRef, Media, OutSlide, Show, Slide, SlideData, Template, Templates, TemplateSettings, Transition } from "../../../types/Show"
+import type { Item, Layout, LayoutRef, Line, Media, OutSlide, Show, Slide, SlideData, Template, Templates, TemplateSettings, Transition } from "../../../types/Show"
 import { AudioAnalyser } from "../../audio/audioAnalyser"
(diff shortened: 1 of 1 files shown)


yurividal:
There is still a different but related bug, which happens when the user has a title slide in the scripture template, and a style overwrite in another output.

The title slide of the secondary output displays the placeholders instead of the metadata, since it cant resolve those values.

Title slide:
image


Title slide in secondary output:
image

vassbo:
Nice, thank you! :)

yurividal:
@vassbo maybe there is still room for improvement. What if instead of the last resort being text[0] what if we always use the style of {scripture_text} as the last resort?

vassbo:
That can work.

vassbo:
Fixed: https://github.com/ChurchApps/FreeShow/issues/2727

PR #2656, opened 2026-01-09
1.5.6-beta.2

vassbo:


src/frontend/components/drawer/media/Media.svelte
```diff
@@ -320,17 +320,19 @@
...
-            if ($activeEdit.items.length) return
             if (activeFile === null || activeFile > 0) activeFile = activeFile === null ? fileCount - 1 : activeFile - 1
         },
         Backspace: () => {
             if (rootPath === path) return
             goBack()
-        }
+        },
+        // macOS workaround as it mostly ignores the M4/M5 mouse buttons
+        // special programs can be used to map these buttons to the following keyboard keys
```
frederickjh: Actually this maps the M4 and M5 buttons to the forward and back shortcut keys on MacOS, so no special program should be needed. See this Apple Support article: [Mac keyboard shortcuts](https://support.apple.com/en-us/102650).

Issue #1925, opened 2025-08-05
Media Folder Cloud Sync


cpaschen:
**Describe the feature**
Add a separate Sync function that synchronizes just the media folder - to a different sync destination (google drive or FTP - per other FR) than the core FreeShow sync location.

This is a follow-up to the discussion in the FreeShow facebook group, and specifically ...

"To sync media on the computer a custom solution will have to be added to the media folder like OneDrive or Google Drive for desktop."

**Additional context**
Although we could set this up manually with each remote system, it would be easier to have this built-in to FreeShow, so that those that need to share media between home and church can do it without extra systems.

BTW ... we have recently been experiencing trouble with having our remote team use OneDrive, possibly due to the fact that we have all 3 OSs (Win, Mac, Linux) used by various team members at home. Keeping all these systems working is challenging. However, the integrated Sync of FreeShow seems to work without any problems.

vassbo:
If media sync would be added I think it would be best to have it in the same system as the existing sync, as users can setup their own system with a different destination if they need to: https://github.com/ChurchApps/FreeShow/issues/619

With that said, I think using the existing cloud sync, we can have a folder called "Media" in the data location (Documents/FreeShow/), and all the contents in that will be synced across devices. If a media file is missing, we can check that location for a match. (Clicking "Bundle all media files" in the settings will move them into the FreeShow/Media folder at the moment, so that would work well)

daves777:
Just wanted to second this request... It would be extremely helpful, as getting media for announcements/backgrounds/etc working between our team members at home has been quite inconsistent and complicated until now. Having all the media live in the freeshow data location and sync between all devices would be so beneficial. Would greatly appreciate this enhancement!

vassbo:
So here's the deal, we're soon going to make our own sync which will work like the Google Sync, just more seamless and easier to set up. It will be free, so we won't be able to have support for media files.

But if there's a way to simplify/streamline the setup of a custom method like OneDrive, Google Drive, FTP, or any computer file system sync that would be nice.

I guess a big help would be to centralize any used media in one selected media folder like I mentioned previously. Meaning if it's taken from that folder nothing will change, but if media is used from another file location it will be copied over to the custom media folder.

Let me know if anyone has ideas for this.

carakas:
Reacted to an old post on reddit but just found this one so I’ll reply here.
We have the same issue, some on windows others on mac.
[...]

PR #2667, opened 2026-01-11
Auto bundle media files on save

shadownetdev1:
- Adds a toggle (defaulting to off) to `Other` that when enabled makes FreeShow auto bundle media files whenever a save happens.
- Fixes `.vscode/settings.json` so that Prettier indents files correctly (as defined on the existing `"editor.tabSize": 4,` line).
-  Adds code to the bundle function that makes FreeShow able to copy media files that are part of a project, but are not in a show.

As this is my first time working on this project I don't fully know what is expected in a MR and I may have made a mistake or done something inefficiently. Let me know if something needs to be changed.

Also not completely sure how I am supposed to handle language files. Do the files need to have an entry that is empty/null? or should they have no entry (like I have currently)? or is there a script I need to run that handles that for me?

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -9,6 +9,7 @@
     "prettier.printWidth": 500,
     "prettier.singleQuote": false,
     "prettier.semi": false,
+    "prettier.tabWidth": 4,
     "typescript.preferences.quoteStyle": "double",
     "typescript.preferences.importModuleSpecifier": "non-relative"
 }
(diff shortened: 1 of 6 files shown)


vassbo:
Thank you! I thought the `editor.tabSize` should have worked which it didn't, but `prettier.tabWidth` makes sense.

You are correct in just modifying `en.json` it will be uploaded to Transifex and the others will get updated once they are translated.

I was just going to re-work the media storage for cloud sync, so I will probably remove the bundle on save, but the project media is nice.

Also the setting should be stored in `settings.json` which is available in the frontend and not `config.json`, but it's okay.

vassbo:
Made some tweaks: https://github.com/ChurchApps/FreeShow/commit/ab34f963883bdaab4cfb3a6a42fb04e53ca0e83b

vassbo:
https://github.com/ChurchApps/FreeShow/issues/1925#issuecomment-3749783329

PR #2694, opened 2026-01-19
fix: Fix can't rename groups

### LINKED ISSUE #2680: Can't Rename Groups
**Describe the bug**
I can't change the name of my groups on my slides
**Reproducing**
right click a slide and select change group - > rename. It pops up the window to do so but it doesn't update the name after you rename.

**Screenshots (Optional)**

**Version (Optional)**
 - OS: Mac
 - FreeShow Version: 1.5.6b3

**Additional context**

Anoian:
Hello, i saw the bug #2680 and i tried to reproduce and i was able to. The bug was straigthforward to fix :)

Closes #2680

### DIFF
--- a/src/frontend/components/helpers/historyHelpers.ts
+++ b/src/frontend/components/helpers/historyHelpers.ts
@@ -440,7 +440,7 @@ export const _updaters = {
 
     show_key: {
         store: showsCache,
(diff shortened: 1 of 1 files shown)


vassbo:
Interesting.. Thank you! :)

Issue #2709, opened 2026-01-21
Grow to Fit bug..?


issachaaar:
**Describe the bug**

I think the "Grow to Fit" feature is bugged to its updated version 1.5.6 

**Reproducing**

I hope to the "Grow to Fit" fix asap.

**Screenshots (Optional)**
  

<img width="614" height="230" alt="Image" src="https://github.com/user-attachments/assets/7f9797fa-bfc3-42d6-b488-0310da4fb90f" />

**Version (Optional)**
 - OS:  Windows 11
 - FreeShow Version:  1.5.6

vassbo:
@yurividal

yurividal:
@issachaaar  can you give me a bit more detail?
I seem to be able to use grow to fit just fine. How are you applying the template?

Can you send a screenshot of what it used to look like on 1.5.5? (Just install 1.5.5 again on top of 1.5.6)

yurividal:
Wait, i was able to reproduce it. I have focussed all my fixes on shrink to fit, and ended up completely forgetting about grow to fix. Let me look into this, and sorry about it.

issachaaar:
Not a problem at all! thank you!!

yurividal:
https://github.com/user-attachments/assets/de55dc9d-185b-412a-bf56-1dfeb33df402

Ok this should be fixed by https://github.com/ChurchApps/FreeShow/pull/2712

vassbo:
Thanks again @yurividal :)

yurividal:
> Thanks again [@yurividal](https://github.com/yurividal) :)

You are welcome. This auto-size thing is becoming quite a challenge. I keep finding edge cases that I never accounted for all the time. Hopefully it's all sorted now.


Perhaps we need some sort of testing suite, or framework to validate each change.

vassbo:
Yeah, I wish it was supported by CSS, that would make everything so much easier.

PR #2710, opened 2026-01-21
Fix grow to fit not working

yurividal:
This fixes the regression introduced on 1.5.6 (https://github.com/ChurchApps/FreeShow/issues/2709) that broke the grow to fit if font-size was bigger than 100. 

Root Cause: 
In `Textbox.svelte`  when `textFit === "growToFit"` was set for text items, the code was always setting `maxFontSize = itemFontSize` (which defaults to 100px). This caused the binary search algorithm to only search between 10px-100px, preventing text from growing larger.


yurividal:
i just noticed that we have the same logic issue on the stagebox and and stage output. let me fix those too

PR #2712, opened 2026-01-21
Fix grow to fit, and stage output auto-size

yurividal:
This fixes the regression introduced on 1.5.6 (https://github.com/ChurchApps/FreeShow/issues/2709) that broke the grow to fit if font-size was bigger than 100.

Root Cause:
In Textbox.svelte when textFit === "growToFit" was set for text items, the code was always setting maxFontSize = itemFontSize (which defaults to 100px). This caused the binary search algorithm to only search between 10px-100px, preventing text from growing larger.

The second commit has further auto-size fixes, that affected stage outputs:

Multiple issues causing text to display at 800px before autosizing, after our previous fix for "grow to fit".

1. STAGE now ignores item.autoFontSize (computed for OUTPUT dimensions)
   and always starts at fontSize=0 to compute its own size

2. Added stageAutoSize to hideUntilAutosized conditions so STAGE 
   properly hides content during autosize calculation

3. Strip font-size from Stagebox container style for autosize items
   to prevent CSS inheritance of 800px MAX_FONT_SIZE

4. Disabled Stagebox's updateAutoSize for slide_text items since
   Textbox handles its own autosize; fixed retry counter to reset
   on slide changes

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -107,7 +107,6 @@
     // remember which item signature we already reset local font size for
     let lastRenderedSignature = ""
     onMount(() => {
-        
         if (preview) {
              // Defer slightly to ensure DOM layout is ready for measurement, preventing 0-width errors
              setTimeout(() => (loaded = true), 20)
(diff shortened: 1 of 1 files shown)


issachaaar:
https://github.com/user-attachments/assets/e1e4ed09-0520-4e29-89d9-7778f72e9e2c

Hello! I watched the video you sent from #2709 and you fixed it already. I checked it from Freeshow today and still doesn't change. Do we need to update our software or do we need to do something to make the "Grow to Fix" bug fix work? thanks!

yurividal:
@issachaaar  you have to wait for the next version to be released, and then update to it. 
Even if it is a beta version. Just wait a few days. @vassbo usually releases a new version every week or 2

issachaaar:
Thank you so much for the updates! Big help!!!!

vassbo:
@yurividal If you convert a scripture to a show, the preview is messed up, but the actual output is fine:
image

yurividal:
> @yurividal If you convert a scripture to a show, the preview is messed up, but the actual output is fine

oh man, here we go again. lol. Ok, i think i know what i missed (again)...

yurividal:
So many test cases... i didn't catch this because my scripture template doesn't use grow to fit.

yurividal:
@vassbo the problem is that the thumbnails currently don't have any auto-size calculations of their own. they just rely on the output auto-size, and scale it down. so, with my recent changes, since the output didn't have its sizes calculated by the time we show the thumbnails, they were defaulting to 800px.
I am implementing a logic where the thumbnails have their OWN auto-sizing, cased on their own container sizes. This should make it more consisent, but there is a small delay when first creating the show, as each slide is being auto-sized. 

I'll submit a patch soon, and you can let me know your thoughts...

vassbo:
Sounds good to me. :)

vassbo:
Here's another one for you, sorry. Using "Grow to fit" on any item with the "List" turned on makes the text super small.

issachaaar:
Hello! I'm just wondering if you have a new updated version of Freeshow? Just out of curiosity, no pressures! Thank you!

yurividal:
> Hello! I'm just wondering if you have a new updated version of Freeshow? Just out of curiosity, no pressures! Thank you!

Did you try the 1.5.7 beta 1?

issachaaar:
Yes! Big help btw. Just excited for the full release after the BETA version.

PR #2714, opened 2026-01-22
enable independent thumbnail font-size calculations

yurividal:
- Remove preview blocking - all contexts (OUTPUT, STAGE, PREVIEW) now calculate independently
- Add previewAutoFontSize field to store thumbnail-specific autosize values separate from OUTPUT
- Thumbnails initialize from their own cache, fallback to OUTPUT cache, then 100px default
- Fix OUTPUT autosize storage blocking - only block temp drawer previews (showId=temp), not OUTPUT

Let me know what you think?


vassbo:
I guess this works.

PR #2715, opened 2026-01-22
Grouped Context UI

vassbo:
Old:
image

New:
image


vassbo:
Old:
image

New:
image

Should be more understandable.

Issue #2606, opened 2025-12-26
Oversize/overflow lowerthird... sometime... and randomly


claudedeslandes:
Once a while, and not all the time, and randomly, our first slide of a show displaying lower third will super enlarge the text... but now all of the time.

Here is the example from the YouTube video.

<img width="1407" height="797" alt="Image" src="https://github.com/user-attachments/assets/4e84cad7-0076-4c25-9721-b2087b04cb02" />

here is the show it self : 

<img width="1190" height="534" alt="Image" src="https://github.com/user-attachments/assets/e67a6b50-12a1-46e0-b1dc-60184f10034b" />

Here is the template : 

<img width="1530" height="878" alt="Image" src="https://github.com/user-attachments/assets/59a71cd1-157f-4340-a30e-c633e6f35972" />

Randomly.  In a mass, the first show's slide overlarge the text.  Then after the mass, we re-started the show, and everything worked fine.

Windows 11, i7

**Version (Optional)**
 - OS: 
 - FreeShow Version: 1.5.5

I think it sometimes happens when the show's first slide is viewed for the first time.... but, works correctly the second time.

vassbo:
@yurividal This didn't happen before.

yurividal:
Does this only happen on the first slide?



>I  think it sometimes happens when the show's first slide is viewed for the first time.... but, works correctly the second time.

This probably means the issue is on the function that applies the template to the show when it is loaded.
What happens if you click the template (or drag it to the first slide) when the issue happens?

yurividal:
Also, I see you have a video or audio associated with that slide. Does it only happen when you have that? Does it also happen when it is a text-only slide?

vassbo:
See this as well @yurividal, it's the same but with only one output, and no template I think: https://github.com/ChurchApps/FreeShow/issues/2616

claudedeslandes:
Good day, .. or goog morning,
     We have a sound track sink with the slide... all of our songs are sink since about 5 or 6 months.  Yes, it only happends (randonly) on the first slide... never on the other slide.  And, if I may add, it only happends when the show plays for the first time, never when we replay the show after the mass for debugging. 

Q1.: This probably means the issue is on the function that applies the template to the show when it is loaded.
What happens if you click the template (or drag it to the first slide) when the issue happens?
A1.: Something to try when I get back to the studio-chapel this week.

[...]

Issue #2729, opened 2026-01-25
Shrink to Fit does not work properly


Ettnamn:
**Describe the bug**
When using shrink to fit in combination with a font size the text will always fill up the whole box and not respect the font size set. In previous versions the logic was that when using shrink to fit it would only change the font size if the text did not fit the text box. But now it seems to be fixed at 100. 

The problem is the same on templates as well. 

**Reproducing**
1. Edit a show/song or a template
2. Choose "shrink to fit" and a smaller font size, lets say 30. 
3. font will not be 30. It will be 100 on the "preview" but in the editor it will be 30. 

If you make a template with shrink to fit option. You can apply the template many time by spamming the template then you will see the show using the correct size. But it will alway go back to 100size font.

**Screenshots (Optional)**

<img width="1635" height="679" alt="Image" src="https://github.com/user-attachments/assets/5d9de63a-bd27-4a66-b021-6e032e1482c0" />

**Version (Optional)**
 - OS: Windows
 - FreeShow Version: 1.5.6

**Additional context**
The problem does not seem to be in version 1.5.5

vassbo:
@yurividal I could also recreate this.

PR #2731, opened 2026-01-26
Fix autosize race condition with font loading

yurividal:
This finally fixes https://github.com/ChurchApps/FreeShow/issues/2606#issuecomment-3798354478

It passed all my tests, but please feel free to test more if you want. 

Cause:
Not enough time for bvrowser to load the elements was causing fallback fonts (Arial/system fonts) before custom fonts (CMGSans-Bold.ttf, CMGSans-Medium.ttf) finish loading. The autosize algorithm measured dimensions with the wrong font, resulting in incorrect calculations.

Example: First load measured 1760x144px with fallback font → calculated wrong size. After fonts loaded and re-render: 1812x196px with correct font → correct size.

Solution: 
- Added `await document.fonts.ready` to ensure custom fonts are loaded before measuring
- Adaptive polling that waits for layout to stabilize (100-150ms initial wait, then 20-50ms intervals, max 500ms)
- Output window gets 150ms initial wait (since its a separate Electron window), preview gets 100ms
- Stops waiting immediately when dimensions stabilize (to avoid unnecessary delays)

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -468,7 +468,7 @@
                 if (line?.text && Array.isArray(line.text)) allText.push(...line.text)
             })
             const itemText = allText.filter((a) => !a.customType?.includes("disableTemplate")) || []
-            let itemFontSize = Number(getStyles(itemText[0]?.style, true)?.[" font-size"] || "") || 100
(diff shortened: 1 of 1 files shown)


yurividal:
The seccond commit fixes https://github.com/ChurchApps/FreeShow/issues/2729

PR #2736, opened 2026-01-26
Fix/autosize font loading race condition

### LINKED ISSUE #2606: Oversize/overflow lowerthird... sometime... and randomly

### LINKED ISSUE #2729: Shrink to Fit does not work properly

yurividal:
This finally fixes #2606 

It passed all my tests, but please feel free to test more if you want.

Cause:
Not enough time for bvrowser to load the elements was causing fallback fonts (Arial/system fonts) before custom fonts (CMGSans-Bold.ttf, CMGSans-Medium.ttf) finish loading. The autosize algorithm measured dimensions with the wrong font, resulting in incorrect calculations.

Example: First load measured 1760x144px with fallback font → calculated wrong size. After fonts loaded and re-render: 1812x196px with correct font → correct size.

Solution:

Added await document.fonts.ready to ensure custom fonts are loaded before measuring
Adaptive polling that waits for layout to stabilize (100-150ms initial wait, then 20-50ms intervals, max 500ms)
Output window gets 150ms initial wait (since its a separate Electron window), preview gets 100ms
Stops waiting immediately when dimensions stabilize (to avoid unnecessary delays)

The seccond commit fixes #2729

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -528,6 +528,22 @@
             console.error(e)
         }
 
+        // smaller in general if bullet list, because they are not accounted for
+        if (item?.list?.enabled) fontSize *= 0.9
+
+        if (item.type === "slide_tracker") {
+            if (cacheKey) writeAutoSizeCache(cacheKey, { signature: cacheSignature, fontSize })
+            markAutoSizeReady()
+            return
+        }
+        // Store in separate field for previews vs OUTPUT
+        if (preview) {
+            if (fontSize !== item.previewAutoFontSize) setItemPreviewAutoFontSize(fontSize)
+        } else {
+            if (fontSize !== item.autoFontSize) setItemAutoFontSize(fontSize)
+        }
+        if (!isDynamic && cacheKey) writeAutoSizeCache(cacheKey, { signature: cacheSignature, fontSize })
+
         markAutoSizeReady()
     }
 


vassbo:
Looks fine to me.

Issue #2661, opened 2026-01-10
Item's order in a template


claudedeslandes:
I created a template made up of two elements: a lower-third text box, and a PNG image with a “ripped” transparent border that’s 60px wide. The whole idea is for the text box to sit behind the PNG, so the PNG can overlap it and create a nice layered visual effect.

The Bible reference is baked into the PNG, and it displays correctly in template edit mode—the image sits on top of the lower-third text box as intended.

<img width="1516" height="731" alt="Image" src="https://github.com/user-attachments/assets/de59a552-9a66-4727-96ad-a670fa7b0e08" />

When viewing the slides, the lowerthird textbox is over the media image.

<img width="508" height="740" alt="Image" src="https://github.com/user-attachments/assets/c1e3fc07-05cb-4003-a3fd-b163dee200c4" />

I have two output : "Le Ciel" and "OBS", .. the media will only show on the OBS output, and that works great... just for your information.  Here is the main output from OBS : 

<img width="1224" height="687" alt="Image" src="https://github.com/user-attachments/assets/d9e8e3ac-badc-4182-b504-45c15cecf943" />

I have tried switching the media in first position and textbox in the second position and vice-versa.  In ether way, I get the same result.

Once again, thank you for FreeShow and all the hours you guys put into it.

Claude :-)

**Version (Optional)**
 - OS:  Windows 11, i7
 - FreeShow Version: 1.5.5 and 1.5.6 beta 2 -->

yurividal:
Have you tried adding the image as an overlay to the template, instead of directly to the template? 

Give it a try

yurividal:
@vassbo i think some other use has just reported this (or a very similar issue) on facebook.
https://www.facebook.com/groups/freeshowapp/permalink/25872711922415746/

vassbo:
I saw that, I tried fixing it a long time ago, but I broke some things in the process.

yurividal:
@vassbo i noticed that this only happens when the text inside the element doesn't change. 
For example, if the user has {scripture_reference}{scripture_name} inside the overlay box, the problem doesn't happen, because the box is redrawn at every slide change. If the content is only {scripture_name}, then the box gets overlayed, because it is not redrawn on the next slide, since it doesn't change. 
I think i know where this code is and i should be able to fix it

vassbo:
Oh nice, thank you for figuring it out. :)

PR #2749, opened 2026-01-27
fix wrong Item ordering on unchanged items

yurividal:
This fixes https://github.com/ChurchApps/FreeShow/issues/2661

Items that had text that didn't change between slides were being marked as static, and therefore were not being redrawn on slide transitions. 
Now, this fix makes it so that:
- Render all items in single loop in original order to maintain order of elements
- Unchanged items render without {#key} wrapper to prevent flashing when there are no animations set
- Transitioning items render with {#key} wrapper for proper animations

### DIFF
--- a/src/frontend/components/output/layers/SlideContent.svelte
+++ b/src/frontend/components/output/layers/SlideContent.svelte
@@ -44,8 +44,9 @@
     let current: any = {}
     let show = false
 
-    // Track which items should skip transition (unchanged content, no real transition)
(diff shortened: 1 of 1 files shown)


yurividal:
This passes all my current tests.
- Autosize grow with and without animations
- autosize shrink, with and without animations
- lists with autosize, with and without animations
- scriptures

vassbo:
Nice, thanks!

Issue #2797, opened 2026-02-04
Style Overwrite not respecting group templates on output.


yurividal:
Opening this bug report for tracking purposes. I'm investigating and working on it. 

<img width="1952" height="979" alt="Image" src="https://github.com/user-attachments/assets/4a4a63db-25cc-4271-8ed8-e0a1cf7611bc" />

<img width="885" height="526" alt="Image" src="https://github.com/user-attachments/assets/53ba47f8-15b2-4e76-8f4b-2f8315e8f20e" />

<img width="850" height="509" alt="Image" src="https://github.com/user-attachments/assets/6efa464a-859b-47cd-929d-f4636030e876" />

I have 3 templates.
Songs, Songs Title, and Songs Copyright

In settings -> Manage Groups I have applied "Songs Copyright" to the ~L group, and "Songs" to the actual verses, and Songs Title to the ~T group.

The template Songs has a style overwrite, that applies a red color to anything within parenthesis, for testing.
The other 2 templates don't have this style overwrite.

When i load a song that has parenthesis in the name and in the song itself, the overwrite is getting properly applied on the preview thumbnails. (only parenthesis within verses themselves are red. parenthesis in title or license slide are not, since they have a different template.)

But, on my actual output, all the slides are getting the overwrite. including the title and license.

PR #2798, opened 2026-02-04
Fix/sync groups to output and template priorities on outputs

yurividal:
This should fix https://github.com/ChurchApps/FreeShow/issues/2797

@vassbo 
Here is what i found:

Group templates (assigned via "Settings -> Manage Groups") were not being synced to output windows, causing all slides to fall back to the show-level template. This resulted in template style overwrites (like parenthesis formatting) bleeding across all slides instead of only applying to their assigned groups.

I don't know since when this has been broken. The fix was simple, but I'm not sure if this was an intentional decision or if this indeed was a bug. Anyways, with this PR, the group templates get synced to all the outputs - Unless the output has a Template Assigned to it too. In that case, that one takes more importance. 

The current template priority with this PR remains:

1) Slide-level template - Manually set on specific slides
2) Output style template - Per-output override in the settings -> styles
3) Group template - Per-group defaults (e.g., different templates for verse/chorus/bridge)
4) Show template - Show-level default
5) Category template

I'm just not sure about 4 and 5, if this is the order you intended this to be.

### DIFF
--- a/src/frontend/components/slide/Textbox.svelte
+++ b/src/frontend/components/slide/Textbox.svelte
@@ -3,7 +3,7 @@
     import { OUTPUT } from "../../../types/Channels"
     import type { Styles } from "../../../types/Settings"
     import type { Item, Slide, TemplateStyleOverride, Transition } from "../../../types/Show"
-    import { categories, currentWindow, groups, outputs, overlays, scriptureSettings, showsCache, styles, templates, variables } from "../../stores"
+    import { currentWindow, groups, outputs, overlays, scriptureSettings, showsCache, styles, templates, variables } from "../../stores"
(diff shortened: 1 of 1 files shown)


vassbo:
Thanks, the category template changes the show template, so we should only need to reference the show template, and not the category template.

yurividal:
So maybe just skip the last commit?

vassbo:
Yes, I thought we had to revert it with another commit?

yurividal:
@vassbo this issue is back on 1.5.8

<del>My style overwrites are now being applied to all the templates, even if the template set by group template doesn't have a style overwrite... </del>

yurividal:
@vassbo ignore me. i was being dumb. its working

Issue #2728, opened 2026-01-25
Cannot delete a show when using cloud sync


mtulen888:
**Describe the bug**
While testing the new cloud sync, I found I cannot delete a show. After deleting + syncing, the deleted show re-appears.

**Reproducing**
Enable the cloud sync. Create a show and sync / close the application.
Delete the show. Resync. The deleted show re-appears.

**Screenshots (Optional)**

**Version (Optional)**
OS: Windows
FreeShow Version: 1.5.6

**Additional context**

vassbo:
I'll try to recreate.

Issue #2746, opened 2026-01-27
Cannot delete 'some' project!


claudedeslandes:
On my home computer, I have a bunch of projects. I managed to clean up and delete about 20 of them, but the last ones won’t delete.

I have two projects: (2026.01.25) and (2026.01.25 2 testing). I can delete the second one, then save and exit—but when FreeShow closes (during the exit/sync process), the project reappears.

Please watch this video to see what I mean.

https://github.com/user-attachments/assets/cc890ccd-6289-4ff6-8be0-87b3a84c68dd

FreeShow 1.5.6, using a good windows 11 machine, B1Church synk.

vassbo:
Related: https://github.com/ChurchApps/FreeShow/issues/2728

hwak456:
Sorry for commenting on this closed issue, but I believe it is still relevant to my case. I am using version 1.5.7 beta2 on windows 11 and stable network. I found that this bug still occurs randomly. Sometimes I can successfully delete projects or shows, but other times they reappear after restarting FreeShow. Note that I have ChurchApps sync enabled.

vassbo:
No worries, I will do some more work on this.

hwak456:
Thanks, I appreciate your hard work.

PR #2800, opened 2026-02-04
fix show delete sync issue

BalogAbel:
## Summary
This change updates the SHOWS merge logic for locally missing shows. Previously, the `created` marker could linger and make a deleted show reappear. Now deletion wins if the current device has already “seen” the show. In ambiguous cases we remain conservative and default to download.

## Changes
- In `syncManager.ts` SHOWS merge, decisions for locally missing shows now compare `lastSeen` vs `cloudModTime`.
- If the device has already seen the show, a missing file means delete (no download).
- If the device has not seen the show, the `created` marker still triggers download.

## Behavior Matrix (When → Then)
- Case A: Show exists in cloud, device has seen it, local file missing → delete (`markAsDeleted`), no download.
- Case B: Show exists in cloud, device has NOT seen it, local missing → download if `created` is set.
- Case C: Show exists in cloud, no `created` marker → delete (`markAsDeleted`).
- Case D: Ambiguous (no `cloudModTime` or no `lastSeen`) → fallback to previous behavior, `created` wins.
- Case E: Offline conflict (both sides changed while offline) → conservative download.

## Why This Is Better
- The previous logic kept `created` until all registered devices had “seen” the show. A stale device could block deletion indefinitely.
- The new rule allows deletion only when the current device has actually seen the show. This matches user expectations: if I already downloaded it and now delete it, it should not come back.
- In ambiguous cases we stay conservative to avoid losing offline work.

## Result
Deleted shows no longer reappear once a device has seen them. Ambiguous or offline conflict cases remain conservative to avoid data loss.

## Tests / Verification
1. A creates a show → B syncs → B deletes → B syncs: the show does not come back.
2. A creates a show → B has never synced: the missing show is downloaded.
3. Offline conflict: A creates, B deletes offline, then syncs: download occurs.
4. Stale `created` marker with a device that already saw the show: deletion applies.

## Assumptions
- The change only affects the SHOWS merge logic in `syncManager.ts` and does not alter data format.
- `lastSeen` is derived from `CHANGES.modified[deviceId]` and is updated on each sync.

Fixes: #2728 #2746

### DIFF
--- a/src/electron/cloud/syncManager.ts
+++ b/src/electron/cloud/syncManager.ts
@@ -175,6 +175,11 @@ export async function syncData(data: { id: SyncProviderId; churchId: string; tea
                             // exists only in cloud
                             const existsLocally = !!localFile
                             if (!existsLocally) {
+                                const lastSeen = CHANGES.modified?.[deviceId] || 0
+                                if (cloudModTime && lastSeen >= cloudModTime) {
+                                    markAsDeleted("SHOWS_CONTENT", fileName)
+                                    return
+                                }
                                 if (isCreated("SHOWS_CONTENT", fileName)) await download()
                                 else markAsDeleted("SHOWS_CONTENT", fileName)
                                 return


vassbo:
The problem was that it was marked as deleted when it was not supposed to, not the other way around, also it would not change the same issue for other files, like projects, only shows got the change!

vassbo:
I realized that there are multiple issues here, but it still doesn't fix the issue with projects, etc.

BalogAbel:
Do you think applying the same approach used for the show sync to project sync and other resource syncs would be a good solution?

vassbo:
I think there's a deeper issue here, I'm investigating. But thanks. :)

PR #2801, opened 2026-02-04
feat: Add tolerance setting for scripture verse splitting

yurividal:
Currently, when "Split long verses" is enabled, verses are split at arbitrary positions near the character limit (whatever space character is closer to the limit), often breaking mid-sentence or at awkward word boundaries. This creates unnatural reading breaks that can disrupt the flow during presentations.

Example: A 192-character verse with size=160 might split like:

"...trees bearing fruit in which is their seed, each [slide break] according to its kind. And God saw that it was good."

Solution
Added a Tolerance setting (0-100) that enables punctuation-aware splitting:

We search for punctuation marks (., ,, ;, :, !, ?) within a window around the split point
Splits at the punctuation for natural sentence boundaries.
Example:
With size=160 and tolerance=20, it searches positions 140-180 for the best punctuation mark
Same verse with tolerance=20:

"...trees bearing fruit in which is their seed, each according to its kind. [slide break] And God saw that it was good."

Other fix included in this PR:
When clicking any part of a split verse, all parts are now automatically selected (instead of just one part). This prevents users from accidentally adding only half a verse to their show (happened multiple times at my church, where the second half of the last verse was missing... lol) . Users can still Ctrl+Click to deselect individual parts if needed.

If Tolerance = 0 , it maintains exact current behavior - no code changes to existing functionality. Users must opt-in by setting tolerance > 0.

Use Cases
Small tolerance (10-20): Find punctuation near the ideal split point
Large tolerance (40-50): Accept splits further from ideal if it means better punctuation placement
Zero tolerance: Original behavior (space-based splitting with rebalancing)

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1441,6 +1441,7 @@
         "verses_on_individual_lines": "Verses on individual lines",
         "divide_long_verses": "Split long verses",
         "split_long_verses_suffix": "Show letter suffix",
+        "tolerance": "Tolerance",
         "version": "Show version",
         "reference": "Show reference",
         "split_reference": "Split reference",
--- a/src/frontend/components/drawer/bible/Scripture.svelte
+++ b/src/frontend/components/drawer/bible/Scripture.svelte
@@ -144,7 +144,8 @@
                     const fullText = verse.getHTML() || verse?.data?.text || ""
 
                     if (isSplit && fullText) {
-                        const splitParts = splitText(fullText, chars)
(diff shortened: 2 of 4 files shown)


vassbo:
Btw. `tolerance` is passed to `splitHtmlText` without being used.

yurividal:
> Btw. `tolerance` is passed to `splitHtmlText` without being used.

Oh yeah  that was my initial attempt to implement this, until I realized it didn't need to be in that logic

PR #2825, opened 2026-02-09
Lock slides inside a show

otonielpv:
In our Church we have some Shows with Slides that should not be DELETED, we use this to trigger some functions to automatize our Stream and our workflow.

This PR adds the possibility to Lock Slides inside Shows, I know there is a function to lock a hole Show, but would be nice to have more flexibility locking some specific Slides to prevent deleting or modifying it.

Example:

image

We have "Imagenes Bienvenida" Show that Contains a Welcome Image (The first Slide) this should not be Deleted, and we add some announcements images all sundays, this last can be modified

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -626,6 +626,7 @@
         "unlink_pco": "Unlink from Planning Center",
         "import": "Import",
         "export": "Export",
+        "save_to_file": "Save to file",
         "import_calendar": "Import calendar",
         "imported": "Imported!",
         "duplicate": "Duplicate",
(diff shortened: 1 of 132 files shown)


vassbo:
Thanks, I guess that's useful. Can we lock slide groups instead of locking individual slides, so lock/unlock by right clicking the group on the right side?! So instead of preventing editing the slide we prevent removing/adding the group to the layout & changing the layout slide, but the slide itself can still be edited, because of the way templates work.

otonielpv:
I believe that if a slide is locked, it should not be possible to edit it in any way, whether through normal editing, templates, or anything else. This is to prevent volunteers or people who are not familiar with the workflow from breaking it.

In our case, we want to prevent the first slide from being deleted or modified under any circumstances.

Right now, templates are being applied to locked slides, but I think I will develop a solution to prevent this from happening and If you want to apply a template to a locked slide you will have to unlock It and apply the template again

What do you think?

vassbo:
Okay, if you can work around templates that would be great, we can still have it applied to groups though, but allow no changes.

otonielpv:
@vassbo done! 😄. If the Slide is locked templates are not being applied, give it a try

PR #2827, opened 2026-02-09
 Alert when importing and no proyect is selected

otonielpv:


vassbo:
Thanks. I think we can instead if no projects is opened trigger the "New project" button to automatically create/open a project, I think that's better when I come to think about it. :)

otonielpv:
But If you already have a project (Like me) and only forgot to click on the project, What is the point on creating a new one? 🤔 

I mean,

I have the project called "Domingo"

image

I tried to import the PDF but It didn't show nothing just because I didn't click on It, no need to create a new project

vassbo:
I see...

vassbo:
Then we can check if there's any project just created and auto open that.

otonielpv:
Thats risky because maybe you will importa a PDF into  a project that you don't want to, so, what about having a list with all the projects and select in which one do you want to import it?

vassbo:
It's fine if they just created the project max a few minutes ago!

otonielpv:
Why the topic of max a few minutes ago? I don't know how people are working on the projects. Do they have many projects created? In our case, we only have one for Sundays.

vassbo:
If someone creates a project an forgets to open it before importing we can auto-open it when it was just created, if not we can create a new blank project! If that's not what the user wants they can copy the files over to another project or import again!

vassbo:
I guess in your case you created the project a long time before importing. Another check we can do is find the latest project and open it if it doesn't have any content.

vassbo:
But you don't have to worry about it, I can do it.

otonielpv:
Well, I think it's much easier to have a list of current projects (not counting archived ones) in case you haven't selected any projects before, and then select where you want to import it from that list in the alert.

vassbo:
This is such a rare thing that we don't need a full list for it, this should help a lot: https://github.com/ChurchApps/FreeShow/commit/00c35c5fc6f88667e10d0a0db89d909e5286e3a9

PR #2837, opened 2026-02-11
feat(transitions): fade through transition

zabackary:
This pull request introduces a new "fade through" transition option, allowing slides to crossfade without a brief delay. While a delay does look better, a crossfade may be wanted by some users (including me). This PR implements that for text output, since media by default crossfades without a delay.

* Adds a `fadeThrough` property to the `Transition` and `TransitionData` interfaces in `src/types/Show.ts`
* Adds a toggle switch to enable/disable this option in the Transition options dialog
* Updates the `en.json` translation file. I'm not sure if this is correct way to do translations, so let me know if it needs to be fixed.

### DIFF
--- a/src/frontend/components/main/popups/Transition.svelte
+++ b/src/frontend/components/main/popups/Transition.svelte
@@ -244,7 +244,7 @@
     // RESET
 
     const DEFAULT_TRANSITIONS = {
-        text: { type: "fade", fadeThrough: false, duration: 500, easing: "sine", custom: {} },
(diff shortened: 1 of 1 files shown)


vassbo:
Nice, the translation is correct. Would it be better to have a e.g. "wait to fade in" duration in percentage set to 50% by default so you can set it to 0% like you want or even 100%..

vassbo:
Or call it "Fade in offset"

zabackary:
Thanks for finishing this off!

PR #2910, opened 2026-02-24
Canva integration

vassbo:
API still not approved, but merging for now (and hiding until approved)

image


o0ZzM4g1c:
> API still not approved, but merging for now (and hiding until approved)
> 
> image

Hi, what has been their answer about the validation delay?

vassbo:
@o0ZzM4g1c Haven't received any reply yet.

PR #2918, opened 2026-02-26
Apply action to all slides in template

otonielpv:
So I want to trigger an action to all the Slides that are being imported from PDF.

So the flow is:

- Create a template with an action
- Set the template to the imported PDF images category
- When Converting the PDF to images all the Slides have the action applied

### DIFF
--- a/src/frontend/converters/imageShow.ts
+++ b/src/frontend/converters/imageShow.ts
@@ -1,9 +1,12 @@
 import { uid } from "uid"
 import type { Slide, SlideData } from "../../types/Show"
 import { ShowObj } from "../classes/Show"
+import { clone } from "../components/helpers/array"
 import { getFileName, removeExtension } from "../components/helpers/media"
 import { checkName } from "../components/helpers/show"
+import { categories, templates } from "../stores"
 import { createCategory, setTempShows } from "./importHelpers"
+import { get } from "svelte/store"
 
 export function createImageShow({ images, name }: { images: string[]; name: string }) {
     // newToast("Creating show...")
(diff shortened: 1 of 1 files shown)


vassbo:
What action do you need on all the slides?

otonielpv:
So I have an action to control OBS and the Camera

vassbo:
But why not use a category action instead? That way it will trigger on any slide, but only once.

otonielpv:
So let me explain myself.

I have this two scenes on OBS

image

This one shows the Camera

image

This shows the Camera + PPT, on the category action of the imported PDF I have an action to switch the scene to the Camera + PPT so the first slide will change the scene in OBS. There are some times that the preacher is not talking about what the PDF shows for a long time so I want people to switch to the Camera only Scene while the preacher is not talking about the PDF.

With applying actions to all slides It prevents from people to forgot to swicth again the scene for those moments that the preacher is actually talking about the PDF.

Not sure if I explained well

vassbo:
Okay, I see that you used the category actions already. The way it currently works is consistent with the template backgrounds and show templates. This will make that inconsistent. It's supposed to be added to all the parent slides. Also it should have referenced the show template, and not the category template.

I think you should instead be able to setup an action that has the "Activate when output changes" custom activation, use the category action to enable the listener (Toggle action), and disable it on last slide or on clear.

otonielpv:
So does the "Activate when output changes" triggers on each slide change?

vassbo:
Yes.

otonielpv:
So I tried but this this implies that people that doesn't know to much about computer have to deal with setting an action to the last slide, my objective is to automatize everything that I can. Why "This will make that inconsistent." The category action only applies to the first slide and If you add an action to a template this action will be applied to all the slides. In my case I will not use the category action, I will use the template only so I want the action to all the slides.

I test to apply normal templates with backgrounds and It worked fine

vassbo:
You can make it disable itself when the output is cleared. The category action will make it enabled again if presented again after it's cleared!

otonielpv:
Let me show the full flow.

This is the category where the converted images from PDFs lives

image

It has a Category Action that triggers this function

image

image

So how can I trigger this function on all th slides of shows that are on PRESENTACIONES? Just PRESENTACIONES NOT OTHER ONES.

I also have other functions to control other categories so I dont want them to be affected.

vassbo:
1. Create a new action that will change the camera back (CAM); set the "Custom activation" to "Activate when output changes"
2. Then in your category action, add a new "action" to "Toggle action" (CAM - On)
3. Then create a third action with "Custom activation" set to "Activate when output/slide is cleared" and add "Toggle action" (CAM - Off) - this third action can probably be on all the time as that does not matter.

PR #2920, opened 2026-02-26
refactor: treat scripture slides as scripture even they are on normal…

otonielpv:
Treat scripture slides as scripture even they are on normal shows so It can Override scriptures with template on styles


vassbo:
Great! Thank you. :)

Issue #1006, opened 2024-11-10
Transitions does not always work


Lpwmack17:
As the title already says, the transitions betwenn slides and media doesn't work, even though everything is activated.

**Version (Optional)**
 - OS: Windows 11
 - FreeShow Version: 1.3.0

https://github.com/user-attachments/assets/c9cedd4a-869e-4685-be3b-e4b4f9c673d8

vassbo:
https://github.com/ChurchApps/FreeShow/pull/854#issuecomment-2391011258

vassbo:
https://github.com/ChurchApps/FreeShow/issues/995#issuecomment-2466830806

vassbo:
Does one of these comments explain your issue?

TPaulManohar:
@vassbo We tried a lot, in recent version(1.3.1), But the transition is not working as expected, Please looking into this. Every time restart the app in the show, that doesn't make a sense.

Even we kept global transitions that also not working.

Lpwmack17:
@vassbo None of these work, tried in previous versions and the recent ones, but in the transitions don’t work as expected or even not work

thebiblelover7:
I've had the same issue

TPaulManohar:
@vassbo we tried in macOS with v1.3.1 transition are working there.. in windows it not working, And what I identify here is when we created a template with added media & text, transition are not working. It get lagging and transition are not working.

vassbo:
As I mentioned this is most likely a Svelte transition bug, that I can't do anything about until I rewrite the entire transition system. Also as I said for some reason it only happens in production and not in the development environment.

PavelS-CZ:
Same issue here, the transitions dont work for me. Ive tried enabling in menu - nothing
enabling in preview - nothing, enabling in edit slide - nothing. :((((

colanct:
We're also experiencing sporadic global loss of transitions (and have been for a long time), as described in [#854 comment](https://github.com/ChurchApps/FreeShow/pull/854#issuecomment-2391011258 ) on Win 10 & 11 PCs.  Restarting the app usually restores transitions.   As the issue occurs less than 10% of the time, we're unable to provide more details.

The most recent occurrence was on 1.3.6 on a Win10 PC.

PavelS-CZ:
FOUND A SOLUTION WORKING FOR ME, So basically when i tried using transitions just while using only **1 MONITOR** it didn't work, but once I used a **2nd  MONITOR** it projected correctly on it.

[...]

PR #2927, opened 2026-02-28
fix(transitions): stabilise existing transition system

Oliver-ke:
### Summary
Fixes five known bugs in the transition system.
No new behaviour is introduced - this is a pure stability pass on three files.

Closes part of #2169. Related: #1006.

---

### `SlideItemTransition.svelte`

- Added a `transitionId` counter so that if `startTransition()` is called
  multiple times in rapid succession (reactive re-runs), only the latest call
  commits state — stale calls bail out early. An `onDestroy` handler increments
  the counter to cancel any in-flight work when the component unmounts. This
  eliminates the rare stuck/broken transition state tracked in #1006.
- Removed the 250ms `outDelay` flash-prevention for media items when the active
  transition type is `none`. The delay served no purpose without an animation
  and caused a visible hold on scripture/image swaps that should be instant.

### `transitions.ts`

- Swapped the Y-axis translate signs for `bottom_top` and `top_bottom` to match
  the established `left_right` / `right_left` horizontal pattern. The vertical
  directions were inverted — selecting `bottom_top` produced top-entry motion
  and vice versa.

### `Background.svelte`

- Replaced the `tooRapid` timer + `tryAgain` boolean with an `isTransitioning`
  flag and a `pendingBackground` slot. Previously, background changes that
  arrived during an active transition were silently discarded (or only retried
  once via `tryAgain`). Now the latest pending change is always queued and
  applied as soon as `loaded()` fires, so rapid slide advances no longer leave
  the background stuck on an earlier slide.

---

## Testing

1. `npm start` — open a scripture show, set transition to **None**, advance
   slides → no black flash
2. Advance slides rapidly (faster than transition duration) → output always
   settles on the correct slide
3. Set transition to **Slide**, test all four directions → `bottom_top` and
   `top_bottom` now enter from the correct sides
4. Advance rapidly through slides with different background images → final
   background is always correct
5. `npx svelte-check --tsconfig ./tsconfig.json` → no new errors

### DIFF
--- a/src/frontend/components/output/layers/Background.svelte
+++ b/src/frontend/components/output/layers/Background.svelte
@@ -25,25 +25,17 @@
     let firstFadingOut = false
     let currentlyLoadingFirst = false
 
-    // WIP changing media while another transitions is not smooth
-    // WIP changing quicly between media might make it not receive updates
-
     let loading = false
     let timeout: NodeJS.Timeout | null = null
-    let tooRapid: NodeJS.Timeout | null = null
-    let tryAgain = false
-    $: if (data) createBackground()
-    function createBackground() {
-        // prevent svelte bug creating multiple items if creating new while old clears
-        if (tooRapid) {
-            tryAgain = true
+    let pendingBackground: any = null
+    let isTransitioning = false
+    $: if (data) createBackground(data)
+    function createBackground(bgData: any = data) {
+        if (isTransitioning) {
+            pendingBackground = clone(bgData)
             return
         }
-        tooRapid = setTimeout(() => {
-            tooRapid = null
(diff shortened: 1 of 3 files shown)


vassbo:
Thanks, I could not see the different behaviour.

`SlideItemTransition.svelte` is remounted each time the show changes, so I can't see how the transitionId makes a difference.
When type is "none" it sets the duration to 0 so the outDelay is not set in that case.

The Bottom to top & Top to bottom is currently correct.

The tooRapid looks to be working as it should (Svelte sometimes makes the item be stuck forever if not). I could not find a case where it didn't advance to the latest slide/background.

PR #2947, opened 2026-03-02
Fix media not adding to empty show

otonielpv:


vassbo:
Thanks for looking into it. Can you remove all the code that was added that does not affect the issue you had, as I guess not all these changes are for the fix?

otonielpv:
@vassbo It was not fixed yet 😆

otonielpv:
@vassbo need this [commit](https://github.com/otonielpv/FreeShow/commit/b3f0fa502424b30ec09f6a88352a1f6368ec10cc) to fix the issue

vassbo:
Oh, sorry. I thought you removed that because it was not related to the fix.

otonielpv:
Yeah I thought the same but, It didn't work 😆

PR #2948, opened 2026-03-02
Granular permissions custimization for user profiles

otonielpv:
Add more permissions to configure for profiles, also fix a behaviour where a profile could change their own permissions, this should be onyl done by the Admin

### DIFF
--- a/src/frontend/components/settings/tabs/Profiles.svelte
+++ b/src/frontend/components/settings/tabs/Profiles.svelte
@@ -34,6 +34,8 @@
     // UPDATE
 
     function updateAccess(key: string, id: string, accessType: AccessType) {
+        if (!canManagePermissions) return
+
         let data = currentProfile.access
 
         let accessData = data[key] || {}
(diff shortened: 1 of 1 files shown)


vassbo:
The Admin can hide the "Profiles" menu if the values should not be changed from the profile.
Also there's no reason to have permissions for media as eveybody should be able to play a video/image/audio/scripture.

vassbo:
Also the Functions should be done on a per tag level and not on a per item level.
But again, anybody should be able to start the available timers/actions.

otonielpv:
@vassbo Functions actions are being done by tags. Variables changed to use tags also, timers and triggers can not have tags.

I have actions and variables to automatize some flows, I don't want people to trigger a wrong action accidentally

vassbo:
We can do tags permissions for Actions/Variables then.

otonielpv:
I just pushed the changes to just add the functions to the profile view

vassbo:
You can open a new PR for that

PR #2954, opened 2026-03-03
Add function permissions to profile

otonielpv:


vassbo:
I don't think we should have permissions for each indivitual timer.

Also instead of `resolveAccessLevel(getAccess("functions"), functionActionTagAccessKey(a.id))` we should do `resolveAccessLevel("actions", a.id)`
Also we can keep the functionsList instead of functionsGroups and just use `resolveAccessLevel(a.id, item.id)` or something.

otonielpv:
done

For the Functions permissions:
Instead of using old key helpers per item, it now uses the simpler style like resolveAccessLevel("actions", tagId)
I also switched from grouped permission data to a simple list approach, as requested.

For timers:
Timers now support tags.
Timer visibility/permission checks are done by timer tags.
New timers can automatically get a tag when created from a tag submenu.

I also updated the related Actions/Variables/Timers screens so they all use the same tag-based permission logic, and kept backward compatibility for old permission keys.

otonielpv:
@vassbo I also added that the Profile tab is allways visible so If a user is running a non Admin profile and I I need to make some changes urgently during the service I don't have to close the app and launch it again to switch the profile.

With the permission disabled  you will have the tab visible but you can not change any permissions

vassbo:
No, if it's hidden you can still access it by pressing arrow down! So that's not needed.

otonielpv:
This doesn't make any sense, so anyone who knows this "trick" could change their own permissions...

Also from the UX perspective you have to know that if you are on settings and press arrow down the profile shows us, why just don't disable the permissions?

vassbo:
Yes, profiles are to organize, and to prevent people that doesn't know what they do from messing something up, if people know what they do they can edit things.

It's the same problem if they can just open the admin profile directly anyway.

vassbo:
I suppose it can be disabled when not admin, but we don't need read only for settings.

otonielpv:
If you have another profile than Admin you should put a password to the Admin one so noone can access the Admin profile.

But with the arrow "trick" someone can press the down arrow on settings by error and display all their permissions and mess it up

otonielpv:
> I suppose it can be disabled when not admin, but we don't need read only for settings.

I did not add an readonly button, the dissable one that was already implemented what it does is prevent the permissions from being edited but the profile tab is allways visible

vassbo:
> I also added that the Profile tab is allways visible

It shouldn't be visible always.

otonielpv:
Why? The arrow down trick is not on the docs, I knew it because you told me, but someone that doesn't know the trick has to relaunch the app to change their profile.

That is not a good UX

vassbo:
Just leave it as visible then, but if it's set to hidden it should be hidden.

vassbo:
image

otonielpv:
Ok, so I'm adding a new read-only button, if it is set to hidden it will be hidden, if set to read-only It will show permission but you can not modify them, is it ok for you?

vassbo:
We don't need the read-only, hidden is hidden, visible is visible.
But only admin can edit profile permissions unrelated to the state.

otonielpv:
Ok, so at least add the arrow down trick to the doc

otonielpv:
@vassbo done, reverted the profile changes and now admin can only edit the permissions

PR #2961, opened 2026-03-03
Implement chord chart parsing and update song section handling

otonielpv:
Songs now pull lyrics directly from Planning Center's chord charts instead of the lyrics section. This preserves the repeat markers (//) that were missing before. The system automatically filters out the chord symbols and keeps only the text clean.

When a section has repeat markers (//), FreeShow automatically creates two identical slides for that section. This lets presenters switch between repeats during live performances without manual editing.

### DIFF
--- a/src/electron/contentProviders/planningCenter/request.ts
+++ b/src/electron/contentProviders/planningCenter/request.ts
@@ -21,13 +21,76 @@ const PCO_API_version = 2
 type PCORequestData = {
     scope: PCOScopes
     endpoint: string
-    params?: Record<string, string> // Add params type
+    params?: Record<string, string>
 }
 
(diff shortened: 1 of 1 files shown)


vassbo:
Okay, great! Please add back the useful comments you removed, and console.error log.

Also the song section object key should still be "lyrics" instead of "chords", as it's mostly lyrics.

otonielpv:
Done, also improved a bit more the Planning Center sequence syncing:

- Now properly detects all section types (FINAL, PRECORO, etc.)
- Allows sections to repeat in your sequence (like CORO3 appearing twice)
- More flexible with spacing and capitalization
- No more empty slides

Song example I used

image

PR #2992, opened 2026-03-07
Allow access to locked slides when removing next timer layout

otonielpv:
We have a slide that is allways locked and we add more slide to add images, we add next slide timer to all of the slides, but if any sunday there are no images extra to add the next slide timer can not be deleted from the locked slide, so project it will move to next section when next timer ends


vassbo:
This makes it inconsistent. Can you use the "Next slide timer" button or action instead?

Issue #3002, opened 2026-03-08
[Remote Show] - Slides with media are all blank on remote show


yurividal:
I have imported a powerpoint, in which every slide is simplt a picture.

When I try to use remote show to view and control freeshow, all the slides in this show are black

On Other PPT, that has text and images, I only see the text. The images are all still black.

otonielpv:
@yurividal did you use an IOS device to connect to the Remote Show?

yurividal:
I tried an iPad, and I also tried using the browser on the own computer that runs Freeshow.

In both cases, all the slides were blank

otonielpv:
Which version are you using of FreeShow? I don't know but maybe it is related with this https://github.com/ChurchApps/FreeShow/issues/3001

yurividal:
1.5.9

I don't think it's related. The entire UI loads just fine.
All normal slides word as expected.
It's only the media, inside slides that have been imported from PowerPoint, that show blank on the remote show screen.

vassbo:
I tested it and the media loaded just fine for me...

vassbo:
Might be fixed by this as it added `encodeFilePath` to the files sent to the remotes: https://github.com/ChurchApps/FreeShow/pull/3005

yurividal:
Sorry guys, I didn't have a chance to test this  as I'm busy with work this week.
But thank you! I'll test it on the church PC whenever the next version comes out

yurividal:
Hey, I just tested this yesterday on my church PC running latest beta3.
The problem still persists. But now the slides are all white.

<img width="1914" height="712" alt="Image" src="https://github.com/user-attachments/assets/2b5040b2-0e3c-437f-82c2-968a4e64f4d5" />

PR #3005, opened 2026-03-09
Add PPT Remote Slide Previews

gladsonsam:
Seems to work, not sure if I broke anything else.

Fixes: https://github.com/ChurchApps/FreeShow/issues/3002


gladsonsam:
Nevermind, now PDF previews are broken.

vassbo:
Ignore the PDFs

gladsonsam:
I don't this this commit breaks the PDF preview, but I can't seem to fix it. I will leave this PR open as it should still fix the original issue but it would be nice if the PDF's could be fixed

vassbo:
Looks good to me, thanks.

PR #3013, opened 2026-03-10
Improve remote PDF slide control and converted PDF handling

otonielpv:
This PR fixes the remote PDF flow, especially for converted PDF shows, and improves stability and slide control behavior across devices.

Main changes

- Improved type detection for converted PDF shows in remote pages.
- Updated PDF thumbnail loading to use converted media pages first, with fallback when needed.
- Switched thumbnail click handling to slide-selection flow for consistent output updates.
- Updated output rendering/handling so converted PDFs are displayed and navigated correctly.
- Restored expected remote controls behavior (active slide highlight, clear actions, previous/next navigation).
- Added translation support for thumbnail size labels.
- Removed an unhandled remote startup request.

Key improvements

- Clicking a thumbnail now reliably changes the live output slide.
- Converted PDFs are handled explicitly instead of relying on fragile assumptions.
- Active slide highlighting now follows real output state.
- Remote behavior is more stable and consistent on mobile and desktop.


vassbo:
I think you didn't checkout from the new dev branch, but okay.

Can you please remove all the extra unneded code. Also we don't need the size options, it can probably be small always for the remote.

otonielpv:
I think I merged the last commit from dev to the branch, also I think the size could help in case someone have lack of sight or want to read something from the remote view

vassbo:
You still got all the commits from the previous dev branch, and not the new dev branch in here, but I suppose it's fine. You probably created the branch from the old dev branch.

The size would be very small on the remote anyway, so you likely would not be able to read it well regardless, but we can make it be medium always then.

You still did not do my request: "Can you please remove all the extra unneded code."

vassbo:
At least make the size default to what makes most sense on mobile/tablet for PDFs and remove the extra size options.

otonielpv:
@vassbo done

vassbo:
Make sure the PDF still works in Safari, after my changes.
Also when it's converted to a show it's no longer a PDF and should not be treated like one.

otonielpv:
No, it is not working, with your changes still crashes

vassbo:
Okay, are you sure you had it as a PDF then, and not as a show when it was working?

vassbo:
So I guess the solution for you is to press "Convert to images", or go to File>Import>PDF. That way it should work.

otonielpv:
No it doesn't work.

What I did:

- Checked out to last commit of this branch (a4322869ac - Cleanup)
- Imported a PDF named ImportedPDFNotWorking.pdf
- Click on the PDF
- Crashes

Here is a video:

https://github.com/user-attachments/assets/697821ed-5db9-4271-ade9-d266d0c21a1e

Then

- Checkout to the previous commit (a17041ef9c - refactor: streamline PDF thumbnail generation by removing debug logging)
- Imported a PDF named (ImportedPDFWorking.pdf)
- Click on the PDF
- It doesn't crash

https://github.com/user-attachments/assets/4f3a1a93-5a63-405e-99de-bc28ab8df272

vassbo:
Okay, I tried just removing the Converted PDF show code you added.
Please either try finding what I did wrong, or create a new PR with just the changes that fixes it and not so much extra, sorry about that. It's better to send multiple small PRs each for one thing, than one that changes many things.

vassbo:
I assume you tried the `ImportedPDFNotWorking.pdf` in the previous commit as well?

otonielpv:
They are the same PDF but with different names

vassbo:
Makes sense, and you added it to the project with drag/drop!?
What if you press "Convert to images", does that work?

vassbo:
My guess is that it's working when it remains as a PDF, but the problem is when it's rendering as a show? In that case it has nothing to do with PDFs, but with media slides.

vassbo:
It makes sense why it would work when you rendered the image show as a PDF instead then.

vassbo:
If that's the case the temporary solution is to drag/drop the PDF into the project so it's actually a PDF, and not imported as images.
And then the fix is something completely different, not related to this PR or PDFs, but to the remote media slide rendering.

otonielpv:
But with drag/drop I can't trigger functions, what we can do is to render the converted pdf to images as a normal PDF on the remote

What do you think? @vassbo 

This will be a small change

vassbo:
As I said earlier: "when it's converted to a show it's no longer a PDF and should not be treated like one."
So when it's a show it should remain a show.

otonielpv:
@vassbo smaller [PR here](https://github.com/ChurchApps/FreeShow/pull/3016) with improved loading performance

PR #3016, opened 2026-03-11
Stabilize Remote Media Slides Behavior

otonielpv:
## Main changes

- **Removed heavy background base64 conversion for Remote shows**
  - Remote no longer bulk-converts slide backgrounds to base64.
  - Backgrounds are now handled through thumbnail paths, which reduces memory pressure.

- **Removed duplicate `SHOW` payload burst**
  - Remote now sends the show payload only once (instead of a quick + full double-send pattern).
  - This reduces unnecessary re-renders and request storms.

- **Added controlled thumbnail queueing for slide backgrounds**
  - Thumbnail requests (`API:get_thumbnail`) are now queued and rate-limited instead of firing in large uncontrolled bursts.
  - All slide thumbnails are queued when opening the show, with throttled processing.

- **Synced slide creation with thumbnail readiness**
  - Slides are rendered in the list only when their background thumbnail (or another directly renderable source) is ready.
  - This avoids creating many empty slide cards first and filling them later.

- **Reduced rendering cost for large slide lists**
  - For large shows, non-active slide cards use a lightweight rendering path (background-focused), while the active slide keeps full rendering.
  - This lowers memory and paint overhead during heavy scrolling.

### DIFF
--- a/src/frontend/utils/remoteTalk.ts
+++ b/src/frontend/utils/remoteTalk.ts
@@ -5,7 +5,7 @@ import type { ClientMessage } from "../../types/Socket"
 import { AudioPlayer } from "../audio/audioPlayer"
 import { loadJsonBible } from "../components/drawer/bible/scripture"
 import { clone, keysToID, removeDeleted } from "../components/helpers/array"
-import { getBase64Path, getThumbnailPath, mediaSize } from "../components/helpers/media"
+import { getThumbnailPath, mediaSize } from "../components/helpers/media"
 import { getAllNormalOutputs, getFirstActiveOutput, setOutput } from "../components/helpers/output"
 import { loadShows } from "../components/helpers/setShow"
 import { getLayoutRef } from "../components/helpers/show"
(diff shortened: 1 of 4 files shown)


vassbo:
Thanks, will check it out later, but looks good.

vassbo:
A few issues:
-

- Firstly the text on the slides does not render until I click on a slide, then just that slide content renders. If I click on another slide it's black again:
image

>
- It just stops rendering at some point (this show has 43 slides):
image

>
- Synced slide creation with thumbnail readiness: We should still render the slides even if the media can't render. It's missing the first slide because of that:
image

>
Other than those issues I think this optimization will help.

This is the working state, it's good to show the broken image when it can't render on the remote just as a visual:
image

otonielpv:
@vassbo Should be fixed, can you try again?

vassbo:
Seems to work fine now, thanks.

otonielpv:
@vassbo seems that with the Cleanup commit has broken again the remote on my iphone

vassbo:
Did it work before the cleanup? I don't think I changed anything.

otonielpv:
Well it didn't break on the clean up it break on 16a2344bc2 - fix(remote): always render slide content with media fallback, I tried and it worked at first but I don't know if it was cached, i'm looking into it

otonielpv:
@vassbo Fixed here [PR](https://github.com/ChurchApps/FreeShow/pull/3021) so on iphone rendering the whole text seems to be what is making the crash also, so I added a special render for iphones with light text

PR #3021, opened 2026-03-12
Fix/remote iphone text only preview

otonielpv:


vassbo:
What was the issue here?

vassbo:
Okay, so the text caused the issue?

vassbo:
Can't we try lazy loading the slides, instead of loading them all at once? That might fix it and we can keep the full formatting like previously.

otonielpv:
So I debugged more and this are the conclussions (with AI explanation):

The original crash is caused by iOS Safari/WebKit being overloaded by rich text preview rendering across many slides, especially during fast scrolling.

This is not about text file size in memory. The expensive part is the rendering pipeline:

1. A large number of DOM nodes per slide preview.
2. Frequent layout/reflow while scrolling.
3. Heavy paint/compositing cost from styled text (multiple spans, effects, transformed/scaled preview surface).
4. Many slide previews active at once, creating frame-time spikes that can trigger tab termination on iPhone.

 The key isolation result was:

- When previews were forced into a lightweight text mode, crashes stopped.
- Therefore, the root issue is the cost of full rich-text preview rendering under scroll pressure, not the data itself.

In short:

The crash is a WebKit rendering saturation issue caused by too many simultaneously mounted rich-text slide previews during aggressive scroll, leading to main-thread and paint/compositing spikes on iPhone.

Lazy-loading improves when content appears, but it does not reduce how expensive each rich text preview is to render. The crash still occurs when WebKit hits a rendering/compositing spike with too many heavy previews active at once.

vassbo:
We should probably have lazy loading of the slides on the remote like we do in the frontend anyways, once it's implemented you can test if that fixes the issue. I think Safari is being overloaded because it's loading too much at once.

otonielpv:
I also tried the lazy-loading, the issue is not the loading the issue is Webkit rendering to much things on the DOM

vassbo:
It does not make sense that it can't load 8 just images though?

otonielpv:
What do you mean by "can't load 8 just images"? The issue with the images on the PDF was different from this, so there were two issues (one is resolved), so now the issue is related to the render of the slides text, even if you click on the show and you load 8 slides and then 8 slides and another 8 slides, the issue is when rendering that slides, so if you scroll really fast the DOM has to render a lot of things in a short time causing the crash. I don't know if I am explaining right. It is not about loading but rendering

vassbo:
Okay, but the problem would be that iPhones would never be able to see the actual preview anymore. Do you think there's a better way, like reducing the scroll speed, and unloading slides out of view.

vassbo:
If there are more than 12 slides.

otonielpv:
So there is a solution: when you are scrolling the slides render as light and when the scroll stops it will render the original slide but only for the visible ones.

I will look also into limiting the scroll speed

otonielpv:
With scroll speed limitation the issue persist so the issue is not the scroll itself, so when I'm scrolling on the slide 15 it crashes because the DOM has to render all this 15 slides, I'm trying to not to render the slides that are not visible

otonielpv:
@vassbo I tried everything, try to render only vissible and hiding the other, limiting the scroll speed but nothing works, I know that with this PR the slides will not be the original ones, but I prefer to have this than nothing

vassbo:
All right, thanks for trying. :)

PR #3030, opened 2026-03-13
Fix delete action not removing groups in shortcuts

otonielpv:
small fix where groups are not being deleted when deleting slides with the del key


vassbo:
The backspace/delete key should not delete slides, but remove groups, which it currently does.

PR #3053, opened 2026-03-17
Feat/pdf import progress feedback

otonielpv:
Add feedback for PDF importing and increase timeout for large PDFs

Last Sunday we imported a 77MB PDF and it was not being imported with no feedback, we reduced the size a lot and it worked.


vassbo:
You don't need to go via the backend to send a message from the frontend to the frontend: sendMain(Main.PDF_IMPORT_PROGRESS in apiHelper.ts

vassbo:
Also the sendToMain(ToMain.PDF_IMPORT_PROGRESS is repeated too much in thumbnails.ts‎

Other than that it looks fine.

otonielpv:
Refactored, the sendPdfImportProgress is called for each imported image to get a real progress also when it is completed and when something went wrong

vassbo:
Thanks, lastly I think we should merge MediaDownloadProgress and PdfImportProgress into a more general progress component, if you don't mind. :)

otonielpv:
Done, now Media downloading and importing PDFs uses the same progress panel 😄

PR #3055, opened 2026-03-17
Add auto-scroll while dragging project items

otonielpv:
Small feature but really helpful, add autoscroll on the borders of the projects area


vassbo:
Autoscroll is already built into Chromium when dragging, and it works!!

I think the issue is the new header actually if that can be worked around (so it scrolls down, but not up because of the header). maybe if we set the header to `pointer-events: none;` when the scrolling is active it should work fine, this applies to the Projects list, the Project content list. It seems to work in the shows list even with the new header.

PR #3056, opened 2026-03-18
Fix/project drag autoscroll

otonielpv:


vassbo:
That won't work as you can't right click the header anymore.

otonielpv:
Reverted the change with only css, now it will set the pointer-events: none; only when dragging

PR #3062, opened 2026-03-20
Fix: make hostname/IP setting per connection instead of global

alanbjoseph:
Fixes global behavior of hostname/IP toggle
Adds per-connection configuration
Allows mixed usage (hostname + IP simultaneously)


alanbjoseph:
Closed due to push to the wrong branch, created another pull request #3065

Issue #1825, opened 2025-07-10
OutputShow is not showing well on version 1.4.7


Eldersteve2:
**Describe the bug**
Hi. I am trying to use an android tablet both as a RemoteShow, and also as a OutputShow. 

The result: The RemoteShow works beautifully!!! A clear and concise description of what the bug is.

OutputShow is extremely lagging behind, and even stops showing images or videos, or text, and the page needs to be refreshed before anything will be displayed. Perhaps there is a better way for me to achieve this???

Please help

Version
 - OS: [Windows]
 - FreeShow Version: [e.g. 1.4.7]

mwb9aa:
Are you using the same tablet to run RemoteShow and Output Show at the same time?

Eldersteve2:
Yes. I tried both, but also by itself. No joy.

*Steve Samuels*


*T: 07851 690 332*
*E: ***@***.*** ***@***.***>*

vassbo:
I have no idea why, but it works completely fine if it's connected on the same computer. But not when connected from another device. And I can't remember changing anything that should affect it.

vassbo:
I got it a little bit improved by not sending any messages until a response is received, it works perfectly fine in a browser on the same computer, it's getting all frames, but for some reason it takes like 5 seconds to receive one message on a different device, and I have not had this issue in the past.

nabsac:
I think for computer to computer streaming over a network there needs to be some image compression.

When using freeshow to stream 1080p window over wifi it was using ~150 Mbps (task manager). This is about 10 times higher usage of my device to device streaming. I notice that as I reduced the output window width and height 720p, 360p, 120p, the video would play smoother. Also, the frame rate increased from 1-2 fps to 10 fps and wifi usage drop to ~24 Mbps. Promising but who is going to watch at 240x120 10 fps.

I ask ai to write me a html5 code for device to device streaming. I had to set up node.js server. It worked really well. It was using compression. So I asked it to modify 3 files in your source code.

src\electron\capture\helpers\CaptureHelper.ts
src\server\output_stream\App.svelte
src\electron\capture\helpers\CaptureTransmitter.ts

[...]

PR #3067, opened 2026-03-21
Optimize OutputShow frame streaming pipeline to reduce latency

alanbjoseph:
### Problem
OutputShow remote browser clients were experiencing ~500ms latency when displaying slide updates from the main Electron window. Investigation identified the lag source was entirely within the capture-encode-stream-render pipeline, with multiple bottlenecks:

- **Frame sampling rate**: Server was capturing at only 10 fps (100ms per frame), vs. 30 fps for other outputs
- **Head-of-line blocking**: Socket.IO transmission was stalled waiting for client acknowledgement before sending the next frame
- **Inefficient frame filtering**: Expensive color conversion was performed on frames that would be skipped as unchanged
- **Client render queue buildup**: Svelte reactive binding could queue multiple pending frames, causing stutter and latency accumulation

### Solution
Optimized the complete pipeline with four targeted changes:

1. **Framerate increase** (CaptureHelper.ts): Raise OutputShow capture from 10 → 30 fps
   - Reduces baseline sampling interval from 100ms to 33ms
   
2. **Backpressure policy** (servers.ts): Replace boolean ack gate with per-output state machine
   - Maintains `inFlight` and `pending` frame state per output
   - Drops old pending frames when new ones arrive (keeps only latest)
   - 250ms timeout recovery prevents pipeline stall on lost acks
   
3. **Conversion optimization** (CaptureTransmitter.ts): Move unchanged-frame skip before color conversion
   - Avoids unnecessary CPU cost on static slides
   
4. **Client render rewrite** (App.svelte): Replace reactive throttle with RAF + latest-frame buffer
   - Event-driven `scheduleDraw()` with `requestAnimationFrame` alignment
   - Single-flight guard prevents render overlap
   - Always displays freshest frame, never queues stale updates

### Expected Impact
- **Baseline latency**: Reduced from ~500ms to estimated <150ms
- **Frame delivery**: Non-blocking pipelined transmission even with slow or laggy clients
- **CPU efficiency**: Skip checks on unchanged frames before expensive conversion
- **Visual smoothness**: Prevents render queue buildup and frame stutter

### DIFF
--- a/src/electron/capture/CaptureHelper.ts
+++ b/src/electron/capture/CaptureHelper.ts
@@ -12,7 +12,7 @@ export class CaptureHelper {
 
     private static framerates: { [key: string]: number } = {
         stage: 20, // StageShow
-        server: 10, // 30 // OutputShow
+        server: 30, // OutputShow
         unconnected: 1,
         connected: 30 // NDI
     }
--- a/src/electron/capture/helpers/CaptureTransmitter.ts
+++ b/src/electron/capture/helpers/CaptureTransmitter.ts
@@ -278,11 +278,11 @@ export class CaptureTransmitter {
         const buffer = image.toBitmap()
         const size = image.getSize()
 
+        if (this.shouldSkipUnchangedNonBlackmagicFrame("stage", captureId, buffer, size)) return
+
         /*  convert from ARGB/BGRA (Electron/Chromium capture output) to RGBA (Web canvas)  */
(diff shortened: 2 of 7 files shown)


vassbo:
There's definetly an issue with the OutputShow, and has been for a while, but only when transfering via the network, it works totally fine when seen on the same device. This PR did not help, in fact changing from 10 to 30 fps made it worse.

Related: https://github.com/ChurchApps/FreeShow/issues/1825

PR #3076, opened 2026-03-22
Add template to planning center sync

otonielpv:
Now everytime there is a planification for planning center, a project is being created with only the songs fo this plannification, we are creating a new project with a specified template every Sundat, copy the songs from the planning center project and deleting it.

This PR adds a dropdown to select a project template that will be applied to all the planning center projects that will be created


vassbo:
I don't get why the items are not just added in the Planning Center service?

otonielpv:
Thats what we are doing right now, copy the song from planning center  project to our Sunday project, but It is much easier to apply the template on the project that planning center creates, so we don't have to create another project (with a template) and move the songs to that new project

vassbo:
What do you have in that template that is not in the Planning Center service?

otonielpv:
Well let me explain better.

We hace a project template that contains all the structure for a Sunday service, every Sunday we create a new project with this template.

Planning Center creates on every planning that you have an empty project with just the songs on it.

The idea Is to apply the project template to the planning center projects, so It already contains the songs an all the structure that we set for Sundays

vassbo:
But isn't the point of importing from Planning Center that you already have the entire structure of the service set up already?

otonielpv:
We are using Planning Center only to manage songs for our worship team.

Our project template has some predefined slides and shows that triggers actions to control OBS, also there are some shows that have Locked Slides.

image

This is our project template, "Imagenes Bienvenida" has an Image that is locked, this also contains a category action, "Inicio y Final" also has a Locked Slide to put the Screen with a black background and trigger another action, "Imagenes Anuncios" also has a category action.

So with all of this structure I don't know if this can be achieved with planning center structure

vassbo:
Okay, I guess. But we only need to show the setting then when Planning Center is connected.

vassbo:
I suppose that's the case already I can see.

otonielpv:
It is already, It only shows when Planning Center is connected

image

image

PR #3114, opened 2026-03-30
Italian language update

ImNightRiderr:
The changes include corrections and improvements to existing translations.


ImNightRiderr (commented):
Hi, I'm an Italian native speaker, and I have updated the Italian localization file (it.json).

This update addresses both corrections and additions: I fixed some existing translations that were inaccurate or inconsistent and added translations for keys that were previously missing.

I hope this was helpful, Thanks and best wishes.

vassbo:
Please join FreeShow on Transifex, and make the changes there: https://app.transifex.com/nettbiter/freeshow/
Because if not it will be overwritten by what's there, also that way you get notified when there's any new strings to translate!

vassbo:
Thanks a lot for your time and service. :)

vassbo:
I see you already joined, I think I managed to upload your changes.

PR #3136, opened 2026-04-05
Add explicit slash for repetitions

otonielpv:
When importing songs from Planning Center, lines with repeat markers (e.g., `//Esto se repite//`) had the `//` stripped from slide content — slides were correctly duplicated but the markers were invisible to the congregation.

## Before / After

```
// Planning Center input:
//Esto se repite//

// Before:
SLIDE 1: Esto se repite
SLIDE 2: Esto se repite

// After:
SLIDE 1: //Esto se repite//
SLIDE 2: //Esto se repite//
```

This will help the congregation to know how many repetitions are

### DIFF
--- a/src/electron/contentProviders/planningCenter/request.ts
+++ b/src/electron/contentProviders/planningCenter/request.ts
@@ -553,9 +553,17 @@ function toSectionSourceLine(line: string): SectionSourceLine {
     }
 }
 
+function addRepeatMarkers(source: SectionSourceLine, text: string): string {
+    if (source.markerOnly) return text
+    const startMarker = source.repeatStartCount ? "/".repeat(source.repeatStartCount) : ""
+    const endMarker = source.repeatEndCount ? "/".repeat(source.repeatEndCount) : ""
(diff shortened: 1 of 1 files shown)


thebiblelover7:
Is this really the intended behavior? Cause in my opinion, I would want to strip the slashes and just have the slide repetitions. With a proper transition, the audience should be able to know to continue repeating the line...

otonielpv:
I've tested it for a couple weeks, and there are people that doesnt do the transition on time, so the congregation told me that it is a bit confusing

thebiblelover7:
Hm... interesting... I found that doing a fade through black that FreeShow defaults to helps queue people to know that a slide changed and that therefore they should repeat it.

I guess this is a specific use case, that I wouldn't use much.

vassbo:
I would do this, like @thebiblelover7:
SLIDE 1: Esto se repite
SLIDE 2: Esto se repite
That's the reason why the default fade shows the text changing even when it has the same content, so people knows it repeats.

Either have the // unchanged from Planning Center, or have it on two slides without //, not both!

PR #3144, opened 2026-04-07
Add Continuous Loop Scrolling Option for Text Items

### LINKED ISSUE #2609: Need loop scrolling effect
Hello,

I’m currently using the scrolling text effect in the Text – Free Show Software app and wanted to share a specific improvement request.

At the moment, when I enter a single line such as:
“For prayer request +710037738”,
the text scrolls from right to left and disappears completely once it reaches the left side. After that, the screen stays empty until the scroll cycle ends.

What I’m looking for is a continuous loop scroll where the same text repeats automatically after it exits the screen, with a configurable gap between repetitions. Ideally, the gap could be equal to 2 to 4 line spaces or a fixed blank distance before the text appears again from the right side.

This kind of seamless repeating scroll would be very helpful for announcements, church displays, and public screens where the message needs to stay visible continuously.

Thank you for your time and for developing this app. I hope this feature can be considered in an upcoming update.

TheFlugeler:
## Summary
Adds a continuous loop scrolling mode to text items, where the text scrolls seamlessly and repeatedly without interruption with a configurable gap between repetitions. This should resolve the issue linked below and allow continuous scrolling for announcements, etc.

Closes #2609

## Changes

**`types/Show.ts`**
- Added `continuousLoop?: boolean` and `gap?: number` to the `Scrolling` interface

**`TextboxLines.svelte`**
- Added a rendering path that duplicates content copies to fill a new scrolling container
- The number of repeated pieces of text is calculated reactively using `bind:clientWidth` and `bind:clientHeight` so it always has enough copies to fill the container regardless of content size
- The distance of the animation is derived from the size of the content size plus the configurable gap, ensuring a seamless loop at a consistent speed
- Existing non-continuous scrolling behaviour is completely untouched

**`boxes.ts`**
- Added `continuousLoop` checkbox and `gap` number input to the scrolling section UI
- Added translation keys for both new fields

## Preview
Scrolling Effect

## How to Test
1. Create a text item on a slide
2. In the item edit panel, open the Scrolling section and select a scroll direction
3. Enable the Continuous Loop checkbox
4. Adjust the Gap value to control spacing between loops
5. Adjust Speed to control scroll rate
6. Preview on the output window

## Notes
- Gap is specified in pixels
- Copy count is calculated automatically based on container and content size, so short content will be duplicated as many times as needed to fill the container


vassbo:
Nice, great! :) It should be continious always without the need for that option.
Also the `<div class="lines"` should probably be extracted into its own component to remove the duplicated code.

TheFlugeler:
I've now made all of the scrolling continuous and removed the `continuousLooping` attribute, but I was unable to extract the `lines` div into a new component. I ran into several errors in chord formatting and dynamically updating values and I didn't want to accidentally remove any functionality that was previously there! I probably don't have enough knowledge of the code base yet to confidently rework some of those functionalities, so apologies if this addition creates more problems than it solves!

vassbo:
Okay thanks. It looks good.

PR #3145, opened 2026-04-09
Fix line-by-line reveal off-by-one and stage display memory leaks

vicholz:
## Changes

This PR fixes two user-facing bugs in FreeShow:

1. **"Reveal line by line" off-by-one**: The last line of a slide was never revealed when using the line-by-line reveal feature, caused by a mismatch between how lines are counted vs how they are filtered for rendering.
2. **Stage display lag/freeze**: The stage display output would get progressively out of sync, lag, and eventually freeze after cycling through several slides/shows, caused by multiple compounding memory leaks and performance issues.

### Modules/Code Changed

**`src/frontend/components/helpers/showActions.ts`** - Fixed `getItemWithMostLines` to use `text.value !== undefined` instead of `text.value.length`, aligning the line counting filter with the rendering filter in `TextboxLines.svelte`. Previously, lines with empty-string text values (e.g., blank lines between verses) were not counted, causing `maxRevealLines` to be smaller than the number of rendered lines, so the last line(s) could never be revealed.

**`src/server/stage/components/Stagebox.svelte`** - Added `onDestroy` lifecycle cleanup for a `setInterval` that was leaking on every component teardown. Since `Slide.svelte` wraps each `Stagebox` in `{#key $stageLayout}` (destroying/recreating on layout changes), each slide change leaked an interval. After cycling through many slides, hundreds of orphaned intervals would overwhelm the event loop.

**`src/server/stage/util/socket.ts`** - Replaced the per-request `socket.on("STAGE", receiver)` pattern in `awaitRequest()` with a single persistent listener and a `pendingRequests` Map keyed by `listenerId`. The old approach added a new listener for every dynamic value request; concurrent requests caused O(n) processing per incoming message.

**`src/server/stage/helpers/show.ts`** - Added `pruneCache()` to cap the `cached` object at 100 entries and remove `isRequested` entries older than 60 seconds. Both data structures were module-level and grew without bound for the lifetime of the page.

**`src/frontend/components/output/Output.svelte`** - Replaced 11+ `JSON.stringify` double-serialization comparisons in reactive `$:` statements with cached-previous-value patterns. Each reactive statement was serializing both the old and new value on every store update; now only the new value is serialized and compared against the cached string.

**`src/frontend/utils/sendData.ts`** - Added a `MAX_SENT_CHANNELS` cap (50) to the `sent` deduplication cache, pruning oldest entries when exceeded. The cache previously retained `JSON.stringify(msg.data)` for every channel indefinitely.

**`src/frontend/utils/listeners.ts`** - Increased the `outputs` store subscription debounce from 1ms to 15ms. The previous 1ms debounce was essentially no debounce, causing a flood of IPC messages to all output windows during rapid slide navigation.

**`src/frontend/components/output/layers/SlideContent.svelte`** - Added a generation counter (`updateGeneration`) to the four-level nested `setTimeout` chain in `updateItems()`. Previously, only the outermost timeout was cancelable; rapid re-invocations could leave inner timeouts from stale calls executing against outdated state.

### New Logic

**Generation counter for nested timeouts** (`SlideContent.svelte`): Each call to `updateItems()` increments `updateGeneration`. Every nested `setTimeout` callback checks `if (gen !== updateGeneration) return` before executing, ensuring stale transition chains from prior invocations are discarded.

**Persistent socket listener with pending-request map** (`socket.ts`): A single `socket.on("STAGE", ...)` listener is registered once via `initRequestListener()`. Each `awaitRequest()` call stores its `{ resolve, timeout }` in a `pendingRequests` Map keyed by `listenerId`. When a response arrives, the listener looks up and resolves the matching pending request, avoiding repeated add/remove listener cycles.

**Cache pruning** (`show.ts`): `pruneCache()` is called on each `replaceDynamicValues()` invocation. It trims `cached` to the most recent 100 entries and removes `isRequested` timestamps older than 60 seconds.

### Breaking Changes

None. All changes are internal bug fixes and performance improvements with no API or behavioral changes for end users.

### Manual Testing Steps
1. Create a slide with multiple text lines including empty lines between verses
2. Enable "reveal line by line" on the text item
3. Advance through the lines using next slide - verify the **last line is now revealed**
4. Open a stage display (web-based or output window)
5. Cycle through 20+ slides rapidly
6. Verify the stage display **remains in sync** and does not freeze or lag

### Automated Tests
No automated tests were added - this project does not have an existing test suite for these components.

## Notes
- The `{#key $stageLayout}` in `src/server/stage/components/Slide.svelte` was evaluated for removal but kept for correctness, since `Stagebox.svelte` has non-reactive initialization (`let itemStyles = getStyles(...)`) that requires full teardown/recreate. With the interval leak now fixed via `onDestroy`, the `{#key}` no longer causes resource leaks.
- The `outputs` debounce was set to 15ms as a balance between responsiveness (smooth visual updates) and preventing message storms. This can be tuned if needed.

### DIFF
--- a/src/frontend/components/helpers/showActions.ts
+++ b/src/frontend/components/helpers/showActions.ts
@@ -109,7 +109,7 @@ export function swichProjectItem(pos: number, id: string) {
 export function getItemWithMostLines(slide: Slide | { items: Item[] }) {
     let amount = 0
     slide.items?.forEach((item) => {
-        const lines: number = item?.lines?.filter((line) => line.text?.filter((text) => text.value.length)?.length)?.length || 0
+        const lines: number = item?.lines?.filter((line) => line.text?.filter((text) => text.value !== undefined)?.length)?.length || 0
         if (lines > amount) amount = lines
     })
     return amount
--- a/src/frontend/components/output/Output.svelte
+++ b/src/frontend/components/output/Output.svelte
@@ -37,8 +37,15 @@
     // output styling
     $: currentStyling = getCurrentStyle($styles, styleIdOverride || currentOutput.style)
     let currentStyle: Styles = { name: "" }
+    let cachedStyleStr = ""
     // don't refresh content unless it changes
-    $: if (JSON.stringify(currentStyling) !== JSON.stringify(currentStyle)) currentStyle = clone(currentStyling)
+    $: {
+        const newStr = JSON.stringify(currentStyling)
+        if (newStr !== cachedStyleStr) {
+            cachedStyleStr = newStr
+            currentStyle = clone(currentStyling)
+        }
+    }
 
     $: alignPosition = currentStyle?.aspectRatio?.alignPosition || "center"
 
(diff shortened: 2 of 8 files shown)


vassbo:
Nice, thanks. Seems good to me.

PR #3162, opened 2026-04-14
Add change output screen action for specific output

otonielpv:


vassbo:
Just have multiple windows in set locations and toggle them on/off. I don't see any need to move one around.

otonielpv:
This is weird, because the toggle action does not have an option for on/off it is just toggle so I can't control when to turn it on and off

vassbo:
We should add a toggle state value then like "Toggle action" does it.

otonielpv:
What do you mean by a toggle? I don't get it

vassbo:
See how the "Toggle action" does it, or "Toggle output lock"

vassbo:
image

otonielpv:
Ah okey, so you mean to add the on/off option to the toggle option windows action?

vassbo:
Yes

otonielpv:
I'm not sure if this will break for the people who are already using it

vassbo:
It won't as Toggle is the default.

otonielpv:
new [PR](https://github.com/ChurchApps/FreeShow/pull/3163)

PR #3167, opened 2026-04-15
Fix planning center sync

otonielpv:
The sync with planning center was not working with this meesage error

image

### DIFF
--- a/src/electron/contentProviders/planningCenter/request.ts
+++ b/src/electron/contentProviders/planningCenter/request.ts
@@ -532,6 +532,12 @@ function extractRepeatDelimiters(line: string): RepeatDelimiterData {
         cleanLine = cleanLine.slice(0, cleanLine.length - endMatch[0].length)
     }
 
+    const endMatchBeforeTrailingChords = !repeatEndCount ? cleanLine.match(/^(.*?)(\/{2,})((?:\s*\[[^\]]+\])+\s*)$/) : null
+    if (endMatchBeforeTrailingChords) {
+        repeatEndCount = getRepeatMarkerCount(endMatchBeforeTrailingChords[2])
+        cleanLine = `${endMatchBeforeTrailingChords[1].trimEnd()}${endMatchBeforeTrailingChords[3].trimStart()}`
+    }
+
     const markerOnly = Boolean((repeatStartCount || repeatEndCount) && !cleanLine.trim())
 
     return { cleanLine, repeatStartCount, repeatEndCount, markerOnly }
--- a/src/frontend/IPC/responsesMain.ts
+++ b/src/frontend/IPC/responsesMain.ts
@@ -338,6 +338,38 @@ export const mainResponses: MainResponses = {
         const replaceIds: { [key: string]: string } = {}
         const allShows = keysToID(get(shows))
         const providerLocalAlways = get(contentProviderData)[data.providerId]?.localAlways ?? false
+        const linkKey = data.providerId === "planningcenter" ? "pcoLink" : data.providerId === "churchApps" ? "chumsLink" : data.providerId === "amazinglife" ? "alLink" : ""
+        const origin = data.providerId === "planningcenter" ? "pco" : data.providerId
+
+        function syncProviderLink(showId: string, providerShowId: string, syncLoadedSlides = false) {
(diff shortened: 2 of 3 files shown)


vassbo:
I'm pretty sure `shows.update` is not needed here as it's done in the `history`.

otonielpv:
You’re right for the linked show case. I moved that path back through the normal tempShows/history flow, so the Planning Center version updates the existing show again when “always use local instance” is off. That also avoids writing to showsCache directly for a linked show that has not been loaded yet, which is what caused the quickAccess error in beta 4.

The only place where I still update shows directly is when we reuse an existing local show by name. That branch does not go through the same history/import path, so the provider link would not be saved otherwise.

I also fixed the repeat parsing case where a line ends with something like // [E]. The repeat is still handled correctly, but the trailing // is now removed before rendering so it no longer shows up in the slides.

vassbo:
I see what you mean now, it did not and should not go through the history as it's not creating a new show. So I guess how you did it the first time was better.

otonielpv:
That makes sense. I updated it to avoid going through history for the linked show case.

It now updates the existing show directly instead of recreating it, so it stays aligned with that flow while still pulling in the latest Planning Center changes when “always use local instance” is off.

vassbo:
But now there are two very similar functions, also `cachedShowsData` is done automatically. I think it's better to use `persistProviderLink` like you did the first time, instead of setting `showsCache` when it does not exist.

otonielpv:
I updated it to reuse persistProviderLink, then loadShows, and only replace the show once the cache entry exists.

I also moved the existing-show sync into setShow, without the extra helper or any manual cachedShowsData update.

vassbo:
I just made the change I wanted, the `shows` vs `showsCache` code is not optimal at the moment.

otonielpv:
But now the changes made in planning center are not applied when syncing againg

So If I do a change on a song on planning center and click on the sync button the show is not being updated with this change

vassbo:
I see, try again now.

vassbo:
And make sure you don't have turned on the "Always use local instance" option.

vassbo:
If you still have the issue, it must be something else unrelated to these changes I think.

otonielpv:
There is something wrong, I start the app, the song loads fine, I do a change on planning center, hit the sync button and the song content disappears

image

vassbo:
I do unload it from the `showsCache` but it should have loaded again, did you try going out from the show and back into it?

otonielpv:
Yes, the only way of getting it right is to shutdown the app and launch it again

vassbo:
I think the issue is that it does not save the updated show to the file, but it looks like that was an issue previously as well.

vassbo:
I didn't account for shows not being saved yet.

vassbo:
But it still does not affect why it didn't update properly in your case.

vassbo:
It should load properly again.

otonielpv:
It loads fine now but it does not update with the planning center changes.

Why not just go back to commit 04c2aef1a5 - Refactor show handling in setShow and mainResponses for improved synchronization, I know it's not optimal but at least it works fine 😅

vassbo:
Are you sure you did not also have the same issue before this PR?

otonielpv:
Before this PR the planning center sync was not even working, I tried the 1.6.0-beta3, the sync works but the changes on planning center are not updating the show too

Also this was not working:

`I also fixed the repeat parsing case where a line ends with something like // [E]. The repeat is still handled correctly, but the trailing // is now removed before rendering so it no longer shows up in the slides.`

vassbo:
From what I can see it was the same behaviour before this PR, so in that case it's a different topic unrelated to the original intent of this PR. Because that specific fix/change was not there when opening the PR for review, you should have opened a new PR for that instead. Or open a draft PR if you want to make multiple different changes before review. (The trailing slashes fix was also unrelated but that was distinct enough). Hope that makes sense.

Looks like it was actually this commit where you made the change: c329e45d16907665159441f768e0384af1a6e671
I got it implemented into dev now, please try it out.

otonielpv:
Alright I will bear that in mind for next time

In other way, with the last commit it's working fine 🎉

Issue #3130, opened 2026-04-04
Project Template Placeholders


nwhsda-dotcom:
**Describe the feature**

**Additional context**

<img width="1752" height="815" alt="Image" src="https://github.com/user-attachments/assets/364c934e-95b2-4266-827d-c39fce0d13ab" />

thebiblelover7:
Is this so that you can add a placeholder to be replaced by a show later on? Or what's your use case?

vassbo:
I think you should use sections as titles instead and have it say e.g. "Songs" if that's what you want.

nwhsda-dotcom:
yes that would be correct. We use the templates for every service but
sometimes miss were songs should be added. this would help with adding
songs into the server / template

thebiblelover7:
Yeah, I think this should be added. If there could be a placeholder show that when clicked prompts to select a show to fill its place, much like ProPresenter does it. So combined with a template, it could say "Song 1" which when you clicked it, could prompt you to select a song for that service. 

Something like this: https://support.renewedvision.com/hc/en-us/articles/40377194830995-Creating-and-Using-Playlist-Templates-in-ProPresenter

PR #3180, opened 2026-04-18
Preserve song positions on planning center re-sync

otonielpv:
I moved the song from planning center to a specific position, syncing with planning center changed all the song positions and moved them to the bottom of the project

### DIFF
--- a/src/frontend/IPC/responsesMain.ts
+++ b/src/frontend/IPC/responsesMain.ts
@@ -2,7 +2,7 @@ import { get } from "svelte/store"
 import type { ContentProviderId } from "../../electron/contentProviders/base/types"
 import type { ToMainSendPayloads } from "../../types/IPC/ToMain"
 import { ToMain } from "../../types/IPC/ToMain"
-import type { Project } from "../../types/Projects"
+import type { Project, ProjectShowRef } from "../../types/Projects"
 import type { Show, Slide } from "../../types/Show"
 import { API_ACTIONS, triggerAction } from "../components/actions/api"
 import { receivedMidi } from "../components/actions/midi"
@@ -338,6 +338,8 @@ export const mainResponses: MainResponses = {
         const replaceIds: { [key: string]: string } = {}
         const allShows = keysToID(get(shows))
         const songOrigin = get(contentProviderData)[data.providerId]?.songOrigin
+        const alwaysUseLocal = songOrigin === "local"
+        const alwaysUseOnline = songOrigin === "online" || songOrigin === "cloud"
         const linkKey = data.providerId === "planningcenter" ? "pcoLink" : data.providerId === "churchApps" ? "chumsLink" : data.providerId === "amazinglife" ? "alLink" : ""
         const origin = data.providerId === "planningcenter" ? "pco" : data.providerId
 
@@ -362,6 +364,16 @@ export const mainResponses: MainResponses = {
             })
         }
 
+        function createOnlineOverwriteShow(show: Show): Show {
+            return {
+                ...show,
+                timestamps: {
+                    ...(show.timestamps || {}),
+                    modified: Date.now()
+                }
+            }
+        }
+
         // CREATE SHOWS
         const tempShows: { id: string; show: Show }[] = []
         for (const show of data.shows) {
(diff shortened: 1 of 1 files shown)


vassbo:
New service items should not be appended to the end, FreeShow should always have the same item & song order that are in Planning Center.

otonielpv:
done, only existing position items will be preserved on resync

vassbo:
You didn't remove the code from the pco-song-origin-sync branch.

otonielpv:
Yeah sorry my fault

vassbo:
Now it looks like it doesn't add changes at all...

otonielpv:
What do you mean? There are changes on responseMain.ts 🤔

vassbo:
If you sync a changed service from Planning Center, it won't change the existing project items.

otonielpv:
I forgot to push the last commit, sorry today is not my day 😔

vassbo:
No worries, I thought you meant you moved the song position in Planning Center when you said; "I moved the song from planning center to a specific position", but I see now that you probably meant that you moved the project item position in FreeShow (as a "preserve FreeShow order" comment was added). So I think it's working as it should already!

FreeShow should use the position from the Planning Center service, that's the point of it syncing again after it has been synced, so if someone re-orders the songs, the FreeShow project should reflect that! If people want a custom position in FreeShow they should duplicate the project instead, so it's not linked to Planning Center.

otonielpv:
Well the issue i was trying to fix is that we are applying a template to our planning center sync, so if a change is made on planning center for a song I have to resync, then all the songs are moved to the botton of the project and I have to reoder them everytime I resync, another option will be to have a single song resync, this could be very useful so you only resync the song that has changed and not all, what do you think?

vassbo:
Then add a source link for those project items from the template, and re-add only those from to the template after the full Planning Center project is synced. Also I guess a single song re-sync would probably not be needed as most people either want to keep all local, or all from online, and it can always just be deleted locally to be re-synced.

otonielpv:
Dind't understand this:

"Then add a source link for those project items from the template, and re-add only those from to the template after the full Planning Center project is synced."

This is our template:
image

We order our songs like this:
image

But If there is any change on planning center and we resync to update the songs, all the songs go back to the bottom of the project like this:

image

And we have to move them again to their correct position

vassbo:
Yes, so if we mark all the template items in the project, we can then re-apply them in the correct position after the new project has been created from Planning Center.

So on second sync, it will reorder the template items relative to the positions they had in the previous project. So if someone swaps the order of "Jonas" and "Me rindo ante ti" in Planning Center, they will be swapped in FreeShow as well, and also be kept correctly relative to the custom project template items. Or duplicate the project to prevent changes from the Planning Center order.

The Planning Center sync is not really made to work with local items as well, but it's possible by doing what I said, if it makes sense.

otonielpv:
What do you mean by "mark all the template items in the project"

To use this as the template? 🤔 

image

vassbo:
Give the template items a unique key in the object so we can differenciate them from Planning Center items.

otonielpv:
Like this?

image

Cancion X are empty shows

vassbo:
No, that's not what I meant. But I suppose once this is added: https://github.com/ChurchApps/FreeShow/issues/3130 it can be integrated into that.

otonielpv:
So how you meant to do it on the other way?

vassbo:
Check the relative positions of the template items vs the Planning Center items in the current project vs the new project.

vassbo:
I added the project template placeholders, and made it so it will replace any placeholders in a set project template first, then append the rest to the end.
I think because of that feature it's probably not be nessasary to do what I described.

vassbo:
Btw. someone said that "[Pre-chorus]" was not pulled in properly from Planning Center, the way [Chorus] & [Verse] is.

otonielpv:
I think the correct key is PreChorus not Pre-Chorus

image

otonielpv:
I've tested the placeholder and they work, thanks!!!

Btw related to the other message, I think the correct key is PreChorus and not Pre-Chorus, maybe there are some issues with special characters on the tags

vassbo:
I see, but we should support "Pre-Chorus" & "Post-Chorus" regardless.

vassbo:
I also don't think PreChorus will be turned into the "Pre-chorus" group we have in FreeShow.

otonielpv:
@vassbo what did he mean with [Pre-Chorus]? Sections on planning Center must be uppercase

vassbo:
That might be a Planning Center thing, but when we import them into FreeShow it should not matter how it's formatted as long as it's in brackets. It should match how you can import any text file with brackets as the group names.

https://www.reddit.com/r/freeshowapp/comments/1sq7ung/freeshow_not_pulling_prechorus_from_planning/

otonielpv:
Oh what I see is that he is talking about sequences, alright, I will try to speak with him on reddit

PR #3181, opened 2026-04-18
Fix planning center song origin sync

otonielpv:
So the Song Origin from planning center optiosn was not working, doing changes on FreeShow and having use allways online instance didn't revert the changes made on FreeShow and updated with planning center, also with the ask option

### DIFF
--- a/src/frontend/IPC/responsesMain.ts
+++ b/src/frontend/IPC/responsesMain.ts
@@ -364,6 +364,16 @@ export const mainResponses: MainResponses = {
             })
         }
 
+        function createOnlineOverwriteShow(show: Show): Show {
+            return {
+                ...show,
+                timestamps: {
+                    ...(show.timestamps || {}),
+                    modified: Date.now()
+                }
+            }
+        }
+
         // CREATE SHOWS
         const tempShows: { id: string; show: Show }[] = []
         for (const show of data.shows) {
@@ -394,7 +404,7 @@ export const mainResponses: MainResponses = {
                     if (globalGroup) slide.globalGroup = globalGroup
                 })
 
-                tempShows.push({ id: linkedShow.id, show: { ...show, origin, name: checkName(show.name, linkedShow.id) } })
+                tempShows.push({ id: linkedShow.id, show: createOnlineOverwriteShow({ ...show, origin, name: checkName(show.name, linkedShow.id) }) })
                 continue
             }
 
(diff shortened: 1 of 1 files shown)


vassbo:
I'm pretty sure this was the cause: https://github.com/ChurchApps/FreeShow/commit/2c7a4084a99f8e41dc8a61408d4e1b407fd54df7

otonielpv:
It's not working, I have always use online instance, made a change on a song on FreeShow, hit resync (It should revert the change on FreeShow and update the song with the content from planning center) and It didn't bring the changes from planning center.

Also tried with the ask option and the popup didn't appear

vassbo:
The popup only appears first time, but if you have it linked it won't show. So don't worry about the Ask option.
Likely this as well: https://github.com/ChurchApps/FreeShow/commit/95e988b183a1e9781f64adbfe8edad027dd2c14f

otonielpv:
Still not working well 😢


https://github.com/user-attachments/assets/c99eb04b-6b9e-420f-b040-2e6c50e371d9

vassbo:
Check that it's actually linked to Planning Center, if it is it should have been pushed into the tempShows array here, double check if it is or not: https://github.com/ChurchApps/FreeShow/blob/dev/src/frontend/IPC/responsesMain.ts#L386

otonielpv:
Yeah it is, I deleted from local and synced from planning center, also there is something weird (unrelated with this) the first and the last song comming from a plannigication in planning center, doesn't have the option unlink from planning center (I tried with a song that has this option, and the behaviour is the same, didn't update with planning center changes)

vassbo:
Thanks, I think I fixed that now.

otonielpv:
Dind't fix it the last commit, still the same behaviour

vassbo:
I meant the new songs from Planning Center should be linked now when you use the "Always cloud" option.
You said the updated song from Planning Center was pushed to the `tempShows` it should have updated then...

otonielpv:
Well I said that the song was linked to planning center, I didn't try the push to tempShows, let me check

otonielpv:
The real issue was the modified date check.

The Planning Center song was reaching the replace flow, but the overwrite was skipped because the local FreeShow song had a newer timestamps.modified.

So the fix was to bypass that modified-date guard for the Planning Center sync path, which makes "Always cloud" overwrite the local song correctly. Here is the [PR](https://github.com/ChurchApps/FreeShow/pull/3186)

vassbo:
Oh I see, thank you. I forgot that the history checked that.

PR #3186, opened 2026-04-20
Fix planning center sync link

otonielpv:


vassbo:
I think this would do it: https://github.com/ChurchApps/FreeShow/commit/ab356c04acac04665d86a0b6b6561725aba65137

otonielpv:
Yeah worked fine now 😄

vassbo:
Nice, thanks for the help. I think there was like three different issues that caused problems. 😅

Issue #3222, opened 2026-05-03
RemoteShow "add to project" misaligned


ianspryn:
**Describe the bug**
When using RemoteShow on a browser, "add to project" button is misaligned when the project is empty.

**Reproducing**
Create a new project with nothing in it.

**Screenshots (Optional)**
Issue:
<img width="1460" height="978" alt="Image" src="https://github.com/user-attachments/assets/53d51dcf-1d82-4c57-be32-4cc3c701d7e3" />

Expected (fixed after adding a show to the project):

<img width="1455" height="982" alt="Image" src="https://github.com/user-attachments/assets/69cd7a9f-1db7-4b0a-93f7-737fdb79a581" />

**Version (Optional)**
 - OS: MacOS
 - FreeShow Version: 1.6.0

PR #3230, opened 2026-05-06
Fix RemoteShow

gladsonsam:
Should fix: https://github.com/ChurchApps/FreeShow/issues/3222

Also should fix an issue I had where edits would revert randomly on RemoteShow before I could save


vassbo:
Thanks!

PR #3231, opened 2026-05-06
feat: prioritize custom group matching and improve slide deduplication logic. Complete spanish translation

dcalonga:
This PR improves how FreeShow identifies song sections during the "New Show" import process. Currently, the system relies heavily on a fixed dictionary, often failing to recognize specific groups that the user has already configured (such as V1, V2, Cr1, Cr2, etc.).

Key Improvements:

Exact Name Matching: When adding a new show, the system now first checks if the bracketed label (e.g., [Cr2]) matches an exact name in the user's custom group list. If found, it automatically assigns that specific group, including its custom color and settings.

Smarter Deduplication: Fixed an issue where sections with identical lyrics were being merged or auto-numbered (e.g., "Cr 1", "Cr 2") even if they were explicitly assigned to different custom groups. This is crucial for songs where the lyrics repeat but the musical arrangement or key changes (e.g., [Cr] vs [Cr2]).

Language Fixes: Completed the translation of several phrases into Spanish that remained in English, ensuring a consistent localized experience.


vassbo:
Thanks looks fine, but too many unrelated changed, so I had to just add the txt.ts file change: https://github.com/ChurchApps/FreeShow/commit/ff7e9ed6c83efddc1000314cb6e5bff5ab001d6c

For the translations you have to do them on Transifex, so they won't be overwritten: https://app.transifex.com/nettbiter/freeshow/

dcalonga:
I understand the reasoning behind keeping the default checks, but I’d like to suggest a different perspective on why the order of these lines is critical for FreeShow's flexibility.

Currently, FreeShow comes with a very limited set of default groups (Intro, Verse, Chorus, etc.). However, many users (especially for live music) need additional sections like 'Post-Chorus', 'Instrumental', or specific tags like 'V1', 'V2', 'Cr2' to handle key changes or stage display directions for musicians.

The main issue is that the line if (get(groups)[group]) return group at the very beginning creates a "hardcoded" behavior. If a user manually creates a group named 'Chorus 2' (because they can have different lyrics and chords between choruses) or 'Post-Chorus' in their settings, the import process will still fail to recognize them automatically because it never reaches the custom name verification logic. It forces users to re-assign groups manually every time they import a new song.

My proposal:
Could we move the line if (get(groups)[group]) return group to the end of the function?

By checking the exact name match against the user's group list first, we allow FreeShow to be truly dynamic.

If the imported label (e.g., [Chorus 2]) matches a name in the user's custom list, it gets assigned automatically.

If no match is found, the function then falls back to the default dictionary and the ID check.

This change doesn't break existing functionality; it simply allows FreeShow to respect the custom environment the user has built.

Regarding the translations, the problem is that some English phrases don't appear in other languages. Simply compare the en.json file to the es.json file and you'll see a difference of more than 15 lines that are present in the en.json file but not in the es.json file. This is why several phrases continue to appear in English, even after the language has been changed.

vassbo:
As I said I already merged your group change, so don't worry about it.

And as I said all the correct strings are present in Transifex and will need to be translated/corrected there.

dcalonga:
I understand the translation workflow via Transifex, so let's focus only on the logic of the group matching function, as it is a data-linking issue, not a language one.

Please look at the attached image. It clearly demonstrates the problem:

_freeshow-names

The Metadata exists: In the right part of the image, you can see I have already created groups named 'Cr', 'P' and 'Fin' with specific colors (red, purple and gold).

The Match Fails: Because the current code checks get(groups)[group] (checking by ID/Key) before checking by Name, the import process fails to link the song section to the existing metadata.

The Result: FreeShow creates new, colorless (grey) groups instead of using the ones I already configured. This forces users to manually fix colors every time they add a song, even if the names match perfectly.

My proposal: By checking for an exact name match against the metadata first, we ensure that if a user already has a 'Cr' or 'P' with a specific color, FreeShow will honor that style immediately upon import.

Could we move the ID check to the end of the function to prioritize these user-defined styles?

vassbo:
I have already added the changes you made, as I'm trying to say!! So it should be there in the next version!!!

vassbo:
@dcalonga Your changes should be live here: https://github.com/ChurchApps/FreeShow/releases/tag/v1.6.1-beta.2

PR #3252, opened 2026-05-13
Update README

gladsonsam:
Thought it would be nice for some change but up to you


vassbo:
Nice, it looks good. Can we change the screenshots as well?

Issue #3191, opened 2026-04-21
Can't Advance to Media with Keyboard in Focus Mode


MyGuyConsultingYou:
**Describe the bug**
 When in focus mode I can't advance from the last slide of one show to a YT video that is immediately next.

**Reproducing**
 Create a project--add a show--add a YT video under the show--go to focus mode--go to last slide in show and arrow or spacebar to next item.

**Screenshots (Optional)**
https://youtu.be/pV9kwa8_xe8

**Version (Optional)**
 - OS: MacOS Tahoe 26.2
 - FreeShow Version:  1.6.0

**Additional context**

PR #3254, opened 2026-05-13
Advance to non-show items in focus mode

jzongker:
In focus mode, next/prev only worked if the next item type was show.  Expanded that to also include video, image and player.

There was a second issue of the video being played in both preview and output and slightly out of sync with each other.   Updated it to only play in the output window.


jzongker:
#3191

vassbo:
Does not fix it properly unfortunately. But the current system is very messy, so I think I want to rewrite the output system to make it simpler and more maintanable.

PR #3274, opened 2026-05-21
Fix RemoteShow logo in production

gladsonsam:
It built fine for me but just check before merging. I think this is the minimal fix


vassbo:
Got this: `Cannot find module '../../../../public/import-logos/freeshow.webp' or its corresponding type declarations.`

gladsonsam:
This works for me when I built it on Windows. Not sure if its the best solution

vassbo:
All right, looks good to me. Thanks! :)

PR #3278, opened 2026-05-22
Improve Canva presentation import and sync

gladsonsam:
Not sure if this is what you wanted, but this allows us to download and display presentation's from Canva in high res.

Does need a bit more work, but let me know what you think


vassbo:
Oh, nice. I think we should avoid the context menu buttons, and have normal buttons instead in the Canva drawer tab, maybe a Convert to show button, and then a Resync button at the bottom left in Canva only shows, similar to the Scripture specific shows...
Also ContentLibraryBrowser.svelte‎ is supposed to be universal and not Canva specific.

vassbo:
Great! Thanks a lot. :)

PR #3310, opened 2026-05-30
Decode Ultimate Guitar entities

gladsonsam:
Fixes lyrics from appearing as:

`We all cry out &ldquo;You are worthy God!&rdquo;`


vassbo:
Nice, thanks! :)

Issue #3038, opened 2026-03-15
Categories and bibles deleted, possible sync issue?


mtulen888:
**Describe the bug**

When starting Freeshow today, all categories and bibles were gone. As far as I can tell, all the song and bible files were still there, with settings_synced.json being replaced by some default(?) version without our categories and bible entries.

A Freeshow auto-backup was created when I first started freeshow today, which I could restore to get the categories back. Though this caused the recent projects and possibly other changes by other team members to be deleted.

Also all media seemed missing from projects at startup, which was caused by Box app (used for mediasync folder) being updated and had to be restarted. This was not strictly a Freeshow error, still it doesn't help for a reliable Freeshow sync mechanism.

**Reproducing**

**Screenshots (Optional)**

**Version (Optional)**
 - OS: 
 - FreeShow Version: 
Windows, Freeeshow 1.5.8

**Additional context**

vassbo:
Was there any new computers synced recently, their "default" categories might have overwritten the existing ones if they chose "Merge" instead of "Download first time only".

Issue #2621, opened 2025-12-31
Help Syncing Show Categories


collinscentral:
I'm setting up cloud sync for the first time. When I synced the data down on my new computer, I lost several show categories (e.g., Scripture, Pro Presenter imports) and, of course, any actions/templates associated with them. Now the categories are missing on both computers, and when I restore from an earlier backup the categories do not return.

I've poked around in the program files a little, and it wasn't obvious to me where the show categories are stored.

thebiblelover7:
@collinscentral Take a look at https://github.com/ChurchApps/FreeShow/issues/2597#issuecomment-3696411151 and let me know if that helps.

vassbo:
Not sure why they got lost. But we will soon have a new and improved cloud sync that should be better.

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.5.6-beta.2

Issue #930, opened 2024-10-25
[Feature Request] Allow full settings to be synced


yurividal:
I'm trying to have the entire AV team to have the same freeShow environment on their own computers.
Currently, i have tried google drive, and also i have tried syncing the full data folder.
But, even with the entire data folder being synced, the Song categories, the templates, and most settings are not synced.

Do i need to sync the %appdata% folder too to accomplish this? 

A simpler way to get full sync would be nice, unless i'm doing something wrong and this is already possible...

yurividal:
Original PC:

![image](https://github.com/user-attachments/assets/cc8421d6-1ac3-4d79-a8f4-c2a3658687f3)


New PC with same data folder, synced from google drive:

![image](https://github.com/user-attachments/assets/f2aeb38f-ca99-47a4-b34f-87c453c8e937)

yurividal:
Thanks. I ended up doing this.
It initially overwrote all my setting on both computers. luckily I had a backup and I was able to replace it.

Maybe, for the future, you might consider a welcome wizard on the first run that allows the user to point to an existing location, and it won't overwrite on the first run?

vassbo:
Great! It should not override anything on first toggle, but the program will automatically take a backup for you when you toggle this so it should be fine!

Issue #834, opened 2024-09-17
[Feature Request] Cloud sync improvements


gkiibzmultimedia:
The cloud sync is working fine, but when there are multiple users working in freeshow, the last one close the app will replace the files in the cloud causing the works from the previous users to be gone.

Is it possible to append the files instead of replacing them when freeshow synching to the cloud?

What is "Media path ID" for in the Cloud setting page?

vassbo:
This is true, but only for changed projects etc. not shows as they are updated by the latest version. The way this works at the moment makes it hard to do as we would need a timestamp tracker on each element. But if you only experienced this issue with projects I might be able to do something about it?

If the "Media path ID" is set to the same on two computers that means the exact media paths are the same (e.g. if an external drive is used), but in most cases they are different that way the different paths will be auto located and stored at the media path location for the current computer.

chanx:
yes, maybe the project settings need improvement, cannot store in one json file. perhaps save to data folder with separate file for separate project? so if one not edit the same project file at least they won't get error. just suggestion.

jerm00:
I was thinking about this the other day, what about a cloud sync using a traditional sync app (I use Nextcloud self host, but google drive, Dropbox, onedrive and etc) so that it monitors changes for you? You wouldn’t need to implement the change monitoring and file locking part at least.

vassbo:
@jerm00 That works for mediam and would work for shows, but the projects etc. are stored in one file, so it won't help on that issue.

gkiibzmultimedia:
> This is true, but only for changed projects etc. not shows as they are updated by the latest version. The way this works at the moment makes it hard to do as we would need a timestamp tracker on each element. But if you only experienced this issue with projects I might be able to do something about it?
> 
> If the "Media path ID" is set to the same on two computers that means the exact media paths are the same (e.g. if an external drive is used), but in most cases they are different that way the different paths will be auto located and stored at the media path location for the current computer.

yes, only with projects.

PR #3335, opened 2026-06-02
Cloud sync overwrites the whole SYNCED_SETTINGS file, deleting bibles/categories the other device doesn't have

lucaxs09:
### Summary
`SYNCED_SETTINGS` (the `settings_synced` store) bundles many independent item-collections — `scriptures` (registered bibles), `categories`, `overlayCategories`, `templateCategories`, `styles`, `profiles`, `timers`, `variables`, etc. During sync it is **not** in `MERGE_INDIVIDUAL`, so it falls into the *full-file replace* path: the newest device's file replaces the cloud file wholesale. Any item the newest device doesn't have — e.g. a bible imported only on the other computer — is **silently deleted from the registry** for the whole team. The `.fsb`/data files survive on disk, but the registry entry is gone, so the bible disappears from the UI.

### Steps to reproduce
1. Team with 2 computers, A and B.
2. A imports bible **X** locally; B imports a different bible **Y** locally.
3. Both sync.
4. Whoever syncs "newest" replaces `SYNCED_SETTINGS` → the cloud ends up with only one side's bibles; the other (X or Y) is removed from the registry on both machines.

### Root cause
```js
const MERGE_INDIVIDUAL = ["OVERLAYS", "PROJECTS", "STAGE", "TEMPLATES"] // SYNCED_SETTINGS missing
...
// replace full files
if (data.method === "replace" || !MERGE_INDIVIDUAL.includes(id)) {
    // ... replaces the entire local store with the cloud file if cloud is newer
}
```
`SYNCED_SETTINGS` is a dictionary of item-collections (like `PROJECTS`), but it's treated as a single opaque file. "Newest file wins" therefore drops every item unique to the losing side.

### The fix
Intercept `SYNCED_SETTINGS` before the full-file replace and **deep-merge** it: union the items from both sides so nothing unique is lost; on a conflicting key, the newer file wins.

Behavior per top-level key:
- **Object-valued keys** (item-collections *and* object settings like `drawSettings`/`scriptureSettings`) are merged **key-by-key**: keys unique to either side are kept, conflicting keys take the newer file. So object settings are no longer clobbered wholesale either.
- **Non-object keys** (scalars / arrays) take the newer file (last-write-wins).

New helper:
```js
function mergeSyncedSettings(localData, cloudData, cloudIsNewer) {
    const merged = clone(localData) || {}
    for (const [collectionKey, cloudCollection] of Object.entries(cloudData || {})) {
        const isCollection = cloudCollection && typeof cloudCollection === "object" && !Array.isArray(cloudCollection)
        if (!isCollection) {
            if (!(collectionKey in merged) || cloudIsNewer) merged[collectionKey] = clone(cloudCollection)
            continue
        }
        const localCollection = merged[collectionKey]
        if (!localCollection || typeof localCollection !== "object" || Array.isArray(localCollection)) {
            merged[collectionKey] = clone(cloudCollection)
            continue
        }
        for (const [itemKey, cloudItem] of Object.entries(cloudCollection)) {
            if (!(itemKey in localCollection)) merged[collectionKey][itemKey] = clone(cloudItem) // cloud-only: add
            else if (cloudIsNewer) merged[collectionKey][itemKey] = clone(cloudItem)             // conflict: newest wins
            // local-only items are kept untouched (never dropped)
        }
    }
    return merged
}
```
Intercept branch (before the `// replace full files` block):
```js
if (id === "SYNCED_SETTINGS" && data.method !== "replace" && !isNewDevice) {
    const cloudIsNewer = await isCloudNewerThanFile(localStore.path, modifiedDates[file.name])
    const merged = mergeSyncedSettings(localStore.store, cloudFileData, cloudIsNewer)
    if (JSON.stringify(merged) !== JSON.stringify(localStore.store)) {
        await safeStoreSet(localStore, merged, id)
        sendMain(Main[id], merged)
    }
    return
}
```

### Known limitation (intentional, minimal scope)
This does not propagate *deletions* of `SYNCED_SETTINGS` items (a deleted item present in the cloud is restored on the next sync). That's the safe trade-off for "never lose data". A complete solution would give these items a per-item `modified`/`created`-`deleted` ledger (like `PROJECTS`); happy to follow up with that as a separate, larger change if you'd prefer that direction.

### Testing
Built from source (1.6.1) on a real 2-team ChurchApps setup:
- Device A had **Bible X** registered locally; the cloud (from device B) had a different **Bible Y**.
- A synced → **both** A and the cloud ended up with **Bible X and Bible Y** (plus every pre-existing bible), nothing overwritten. With the stock client, one of the two was deleted.
- Unit-level checks of `mergeSyncedSettings` (union, newest-wins conflict, local-only preserved, object settings merged key-by-key) all pass.

### Related
- #3038 *Categories and bibles gone after sync (`settings_synced.json` replaced by default)*
- #2621 *Show categories lost when syncing*
- #930 *[Feature Request] Allow full settings to be synced*
- #834 *[Feature Request] Cloud sync improvements — add files instead of replacing*


vassbo:
What if a value like a bible is deleted, then it would not be removed?

vassbo:
I think having modified for each key would be too much.

But this shouldn't be an issue since the changed settings should be automatically sent over in real time to any connected computer by the cloud socket.

vassbo:
Maybe the socket didn't work correctly?

lucaxs09:
Good point — and that's actually the known limitation I flagged in the PR description: the plain union doesn't propagate deletions, so a deleted bible would come back.

A cleaner way to handle it, without adding a per-key `modified`: reuse the existing `created`/`deleted` ledger in `changes.json` — the same one `PROJECTS` and bibles already use — applied per-item to the `SYNCED_SETTINGS` collections. A deleted item gets marked in the ledger and stays deleted, so deletions propagate correctly with no extra timestamps.

```js
// changes.json already tracks per-device creations/deletions, e.g. for shows:
{
  "devices": ["deviceA", "deviceB"],
  "created": { "SHOWS_CONTENT_My Song.show": ["deviceA"] },
  "deleted": {}
}

// Same idea, applied per-item to the SYNCED_SETTINGS collections.
// Deleting a bible locally would record it in the ledger:
//   markAsDeleted("SYNCED_SETTINGS", "scriptures/<id>")
//   → "deleted": { "SYNCED_SETTINGS_scriptures/<id>": ["deviceA"] }
// On the next sync, an item marked as deleted isn't re-downloaded (the deletion propagates),
// while an item marked as created still gets downloaded.
```

On the socket: you're right that it works well when both machines are connected at the same time. But the data-loss case I hit is asynchronous — one machine imports a bible while the other is closed/offline. The socket never delivers that change, so when the second machine opens and syncs, the snapshot merge is the only thing that runs, and that's where it's lost. (My repro was exactly that: manual sync, one machine at a time.) The socket and the merge cover different situations.

In short: treat `SYNCED_SETTINGS` like `PROJECTS` (both are dictionaries of sub-collections). Happy to open this as a new PR against `dev` if it sounds good to you.

vassbo:
> you're right that it works well when both machines are connected at the same time. But the data-loss case I hit is asynchronous — one machine imports a bible while the other is closed/offline.

There's the problem. The offline machine should always download the full settings file then from the cloud when set to merge if it has newer updates (regardless of individual keys). But it looks like it does not always do that at the moment, maybe it receives a newer update when opening before syncing?

lucaxs09:
Thanks for looking into it. The startup sync itself works fine — the offline machine does download the cloud file on launch. The bug is in the replacement step: when settings are merged, the most recently modified version wins and replaces the **whole file**, regardless of individual keys. So whichever device synced its settings last overwrites everything else in the cloud — including items it never had, even though nothing was actually deleted.

My goal here is simply a sync we can fully trust — zero data loss. The fix is the per-item ledger from my earlier comment: treat `SYNCED_SETTINGS` like `PROJECTS`, using the `created`/`deleted` entries already in `changes.json`, so an item another device just doesn't have is never dropped (only ones actually marked `deleted`). Happy to open a PR against `dev`.

lucaxs09:
For context on how we hit this: we often use machines offline — events, or venues with no internet — making changes during a whole service and only syncing once back online. On reconnect, if others changed the cloud in the meantime, the previously-offline machine replaces their changes with its own settings file (its last change is the most recent). The longer it was offline, the more is lost at once — silently.

Easy to reproduce on the official `1.6.2-beta.1`:

1. Device A connected to the cloud; Device B disconnected.
2. On A, add a few settings-level resources (bibles, categories, variables…) and sync.
3. On B (offline), make one minimal change — e.g. add a single empty category.
4. Connect B and sync → everything A added that B didn't have is gone, and it then propagates back to A.

And that's just one machine — let alone a prolonged server outage, which would push every device into this same state at once.

vassbo:
Oh, because it's opened but not online. If the program was closed while offline this would not be an issue.
So if the machine is always online when opened it should be fine. I agree that we want zero data loss.

The easiest solution would be to prevent changing certain things (the keys in the synced settings) if cloud sync is turned on, but the device is offline. So tell people that they need to connect in order to add a scripture, not the best, but certainly the easiest.

> A cleaner way to handle it, without adding a per-key modified: reuse the existing created/deleted ledger in changes.json — the same one PROJECTS and bibles already use — applied per-item to the SYNCED_SETTINGS collections. A deleted item gets marked in the ledger and stays deleted, so deletions propagate correctly with no extra timestamps.

I would gladly accept this solution as it sounds like the best one, if you would like to take on the task. :)

lucaxs09:
Sounds great — thank you! I'll take it on. I'll work on the per-item ledger approach (treating `SYNCED_SETTINGS` like `PROJECTS`) and open a PR against `dev` once I have everything well tested. 🙏

lucaxs09:
Quick question before I open the PR: I noticed the sync module doesn't have unit tests yet. Since sync is exactly where data-loss bugs live (like this one) and it's easy to regress with future refactors, I think it'd be really valuable to guard it with tests.

I've actually already put together a small **vitest** suite for the per-item merge logic — including a **regression test that reproduces this exact bug**: the old whole-file overwrite drops another device's items, while the per-item merge keeps them. It's pure/isolated (no app launch needed), so it's fast and CI-friendly.

Would you be open to adding vitest to the project? I'd keep it minimal and in a separate PR from the fix. Totally your call on the tooling — happy to adapt to whatever setup you'd prefer.

PR #3354, opened 2026-06-06
fix(sync): merge SYNCED_SETTINGS per-item via the created/deleted ledger

lucaxs09:
Fixes the data loss from #3335: `SYNCED_SETTINGS` was synced as a whole-file overwrite (newest wins), so bibles/categories a device didn't have were dropped from the registry when it synced.

This treats `SYNCED_SETTINGS` like `PROJECTS`: its item-collections (scriptures, categories, styles, …) are merged **per-item** via the existing `created`/`deleted` ledger in `changes.json`, so items unique to either device survive and deletions still propagate — no per-key `modified` needed. Atomic settings (drawSettings, scriptureSettings, …) keep the previous newest-file-wins behavior.

Approach discussed and agreed in #3335. Validated e2e across two devices (offline + online: union without loss, and deletion propagation).

**Follow-up (ready):** I've also prepared a separate PR that extracts the per-item merge into a small pure module and adds **14 unit tests (vitest)** for it — including a **regression test that reproduces this exact bug** (whole-file overwrite drops the other device's items; per-item merge keeps them). I'll open it once you confirm vitest is welcome (asked in #3335) — kept out of this PR to keep the fix minimal.

### DIFF
--- a/src/electron/cloud/syncManager.ts
+++ b/src/electron/cloud/syncManager.ts
@@ -61,7 +61,13 @@ async function deleteLocalFiles() {
 const DEBUG_MODE = false && !isProd
 
 const EXTRACT_LOCATION = path.join(app.getPath("temp"), "freeshow-cloud")
-const MERGE_INDIVIDUAL = ["OVERLAYS", "PROJECTS", "STAGE", "TEMPLATES"] // "EVENTS", "THEMES"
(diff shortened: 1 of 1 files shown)


vassbo:
Awesome! Thanks. :)

PR #3355, opened 2026-06-06
test(sync): unit tests + pure module for per-item SYNCED_SETTINGS merge

lucaxs09:
Follow-up to #3354 (the fix) — adds unit tests for the per-item `SYNCED_SETTINGS` merge.

It extracts the per-item ledger decision logic from `syncManager` into a small **pure, dependency-free module** (`SyncLedger`) — no module globals, no I/O — and `syncManager` delegates the `SYNCED_SETTINGS` collection merge to it. Then adds **vitest** with **14 unit tests**, including a **regression test that reproduces #3335** (whole-file overwrite drops the other device's items; the per-item merge keeps them).

Run with `npm run test:unit`.

Opened as a **draft** because it adds vitest as a dev dependency — I asked in #3335 whether unit tests / vitest are welcome. Totally happy to adapt the setup (or drop it) to whatever you prefer.

Note: stacked on #3354. Until that merges, the diff here also includes the fix commit; once #3354 lands on `dev` this will show only the tests + refactor.

### DIFF
--- a/config/testing/vitest.config.ts
+++ b/config/testing/vitest.config.ts
@@ -1,0 +1,13 @@
+import { fileURLToPath } from "node:url"
+import { defineConfig } from "vitest/config"
+
+// Unit tests live alongside source as *.test.ts under src/.
+// The Playwright e2e (config/testing/start.test.ts) is excluded — it uses @playwright/test.
+// This config sits in config/testing/, so point the root back at the project root.
+export default defineConfig({
+    root: fileURLToPath(new URL("../../", import.meta.url)),
+    test: {
+        include: ["src/**/*.test.ts"],
+        environment: "node"
+    }
+})
--- a/package-lock.json
+++ b/package-lock.json
@@ -102,6 +102,7 @@
                 "tslib": "^2.8.1",
                 "typescript": "^4.9.5",
                 "vite": "^4.5.14",
+                "vitest": "^2.1.9",
                 "xml2js": "^0.6.2",
                 "youtube-player": "^5.6.0"
             },
@@ -177,6 +178,7 @@
                 }
             ],
             "license": "MIT",
+            "peer": true,
             "engines": {
                 "node": ">=18"
             },
@@ -198,6 +200,7 @@
                 }
             ],
             "license": "MIT",
+            "peer": true,
             "engines": {
                 "node": ">=18"
             }
@@ -643,7 +646,6 @@
             "dev": true,
             "license": "BSD-2-Clause",
             "optional": true,
-            "peer": true,
             "dependencies": {
                 "cross-dirname": "^0.1.0",
                 "debug": "^4.3.4",
@@ -665,7 +667,6 @@
             "dev": true,
             "license": "MIT",
             "optional": true,
-            "peer": true,
             "dependencies": {
                 "graceful-fs": "^4.2.0",
                 "jsonfile": "^6.0.1",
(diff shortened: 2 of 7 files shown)


vassbo:
Okay, cool. But can we have the test `vitest.config.ts` file in the `config/testing/` folder?!

`SyncLedger` is currently a duplicate of the existing logic? I suppose it is to not break the existing logic, but the new `requireModifiedTime` key is not used anymore in `checkCloudEntry`.

lucaxs09:
Thanks for the review — you're right on both counts.

1. **`vitest.config.ts`** → moved to `config/testing/` (next to `playwright.config.ts`); the npm script points there now.
2. **The duplication** — good catch. `SyncLedger` is now the single source of truth: `checkCloudEntry`/`checkLocalEntry` delegate to `resolveCloudEntry`/`resolveLocalEntry`, and I removed the duplicated ledger helpers from `syncManager` along with the now-dead `requireModifiedTime` parameter. The per-item collection merge and the modified-time tie-break for regular entries are unchanged in behavior — just no longer duplicated.

Rebased onto current `dev` now that #3354 landed, so the diff is just the tests + refactor.

**Tests:** `npm run test:unit` → **15 passed** · `tsc` clean · prettier clean.

<details><summary>test breakdown</summary>

```
SyncLedger — per-item merge (#3335)
  union (items unique to a device survive)
    ✓ a brand-new local item is uploaded and marked created
    ✓ an item created by another device is downloaded, not dropped
    ✓ two devices adding different items both keep theirs (no whole-file overwrite)
  deletion propagates (a deleted item must not come back)
    ✓ marked deleted but still present locally → deleted (propagated)
    ✓ marked deleted and missing locally → skipped (no resurrection)
    ✓ on both sides and not deleted → kept (caller tie-breaks by date)
  tombstone vs re-create
    ✓ re-importing a tombstoned item by a device that didn't delete it → dropped
    ✓ restoring an item the device itself deleted → revived
  ledger mechanics
    ✓ does not track changes with a single device
    ✓ marking created clears a prior deleted mark (and vice versa)
    ✓ drops the entry once every device has acknowledged the change
  new device adopts everything
    ✓ creates cloud items and never deletes when isNewDevice
  regression: the #3335 bug
    ✓ OLD behavior (whole-file overwrite) LOSES the other device's item
    ✓ NEW behavior keeps BOTH devices' items — bug fixed
    ✓ NEW behavior still propagates a real deletion (doesn't blindly union)

Test Files  1 passed (1)
     Tests  15 passed (15)
```
</details>

Let me know if you'd like anything else adjusted!

vassbo:
Okay, it looks good now. But can you test that the other config files like shows/projects/stage are still replaced correctly with the new system, if you haven't?

For some reason in the current live version a show is randomly not there (does not load from the project), not sure why that happens.

vassbo:
Just merging for now, let's hope it does not break something.

vassbo:
@lucaxs09 Please try out the syncing: https://github.com/ChurchApps/FreeShow/releases/tag/v1.6.2-beta.2

lucaxs09:
Will do — I'll give it a proper test later today (in ~8 hours or so) and let you know how it goes. 🙏

lucaxs09:
Also — I went ahead and wrote the unit tests for the shows/projects/stage path you mentioned (the newest-by-`modified` resolution). Opened them in a small follow-up PR: #3365 🙏

lucaxs09:
I can't run the full e2e sync test right now, so for the moment it's just the unit tests (#3365). I'll do the real two-machine run and report back here as soon as I can.

Thanks again, cheers! 🙏

PR #3357, opened 2026-06-06
Fix Spotify controller on Linux (controls, seek, progress bar)

### LINKED ISSUE #3356: Spotify widget controls not working on Linux
The Spotify controller widget works fine on Windows, but on Ubuntu the controls don't work for me:

- Play/pause, next and previous buttons do nothing.
- Clicking/dragging the progress bar doesn't change the position.
- The progress bar stays at 100% (the left side shows the full duration).

The song info (title, artist, album art, duration) displays correctly — only the controls and progress don't work.

Same result on two Ubuntu machines:

- **Machine A** — Ubuntu 26.04, GNOME/Wayland, Spotify 1.2.86 (native), FreeShow 1.6.2-beta.1
- **Machine B** — Ubuntu 25.10, GNOME/Wayland, Spotify 1.2.90 (native), FreeShow 1.6.2-beta.1

Is anyone else seeing this on Linux? I couldn't find an existing issue for this, so I looked into it and found some possible fixes. I've opened a draft PR (#3357) with the changes — it's working on both machines above. If someone can confirm the bug, happy for it to be reviewed and merged.

lucaxs09:
Fixes the Spotify controller on Linux. All three issues are in the `dbus-send` usage in `src/electron/utils/spotify.ts`:

1. **Controls did nothing** — commands were sent via `dbus-send` without `--type=method_call`. `dbus-send` defaults to `signal` (per its man page), which the player ignores (exit code 0, no effect). Reading state worked only because it uses `--print-reply`, which implies `method_call`. → added `--type=method_call`.
2. **Seek errored** — `SetPosition` used `x64:` as the argument type, which isn't a valid `dbus-send` type name (`x` is the D-Bus signature code; the `dbus-send` name is `int64`), failing with `dbus-send: Unknown type "x64"`. → changed to `int64:`.
3. **Progress bar stuck at 100%** — position was parsed with `/uint64\s+(\d+)/`, which matched `mpris:length` (the first `uint64` = duration) instead of `Position`, so `positionSec === durationSec`. Also, the MPRIS `Position` property is `int64`, not `uint64`. → changed to `/\bint64\s+(\d+)/`.

Tested on Ubuntu 26.04 and 25.10 (GNOME/Wayland, native Spotify client): play/pause, next/previous, seeking and the progress bar all work now.

Closes #3356. Opening as draft until the bug is confirmed.

### DIFF
--- a/src/electron/utils/spotify.ts
+++ b/src/electron/utils/spotify.ts
@@ -126,7 +126,7 @@ export async function getSpotifyState(): Promise<SpotifyState | null> {
                     const artist = out.match(/string\s+"xesam:artist"\s+variant\s+array\s+\[\s+string\s+"([^"]+)"/)?.[1]
                     const len = parseInt(out.match(/string\s+"mpris:length"\s+variant\s+uint64\s+(\d+)/)?.[1] || "0")
                     const art = out.match(/string\s+"mpris:artUrl"\s+variant\s+string\s+"([^"]+)"/)?.[1]
-                    const pos = parseInt(out.match(/uint64\s+(\d+)/)?.[1] || "0")
+                    const pos = parseInt(out.match(/\bint64\s+(\d+)/)?.[1] || "0")
                     const vol = parseFloat(out.match(/double\s+([\d.]+)/)?.[1] || "1")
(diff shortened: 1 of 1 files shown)


vassbo:
Great, thanks! I never tested it on Linux.

PR #3360, opened 2026-06-07
Fix Windows output visibility during Aero Peek

gladsonsam:
Lots of PR's this week but I'm fine if this one gets closed, more so to give you an idea.

Currently, even with always on top enabled, the output gets interrupted during Aero Peak (when I hover on an app in the taskbar). This gets annoying especially when I have multiple windows of an app open, as hovering on any one of them makes all outputs go to the background which gets distracting. I tried screensaver method, but it seems that aero peak is implemented at a lower level, so this uses a hacky PowerShell method.

Would be great if this could be fixed, but there might be a better fix.

https://syobochim.medium.com/electron-keep-apps-on-top-whether-in-full-screen-mode-or-on-other-desktops-d7d914579fce

### DIFF
--- a/src/electron/output/helpers/OutputAlwaysOnTop.ts
+++ b/src/electron/output/helpers/OutputAlwaysOnTop.ts
@@ -1,0 +1,66 @@
+import { execFile } from "child_process"
+import type { BrowserWindow } from "electron"
+
+const OUTPUT_ALWAYS_ON_TOP_LEVEL = process.platform === "win32" ? "screen-saver" : "pop-up-menu"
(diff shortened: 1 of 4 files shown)


vassbo:
Interesting, I also have found that annoying, but didn't know about an easy way to prevent it.
This seems relatively straight forward as long as it's reliable..

vassbo:
I think it's good, thank you.

gladsonsam:
Searched around for a bit but couldn't find anything yesterday, I guess this works then

PR #3379, opened 2026-06-14
Toolchain modernization (Svelte 5 / Vite 8 / TS strict), dependency + security updates, and a unit-test suite

jcfurey:
A batch of tooling, dependency, and quality work from my fork.

- **Toolchain:** Svelte 5 + Vite 8 + TypeScript 5 (compat mode); ESLint 9 flat config.
- **Dependencies/security:** Electron 37→40, music-metadata 7→11 and other advisory clears; safe replacement for `new Function()` math eval.
- **Type safety:** full TypeScript `strict` across frontend/electron/server (0 errors); cleared 78 Svelte 5 migration errors; fixed a few latent bugs (e.g. AudioFilter Float32Array EQ curve).
- **Tests:** vitest suite (251 tests) covering core utilities and all 19 import converters incl. ProPresenter (base64/RTF); jsdom for XML formats.
- **Build:** skip macOS notarization when Apple creds aren't set so local `npm run pack` works.
- **Docs:** BUILDING.md, ACCESSIBILITY.md, Dockerfile.

I realize this is a rather large and mixed-work PR. I am more than happy to split into more focused PRs if requested/warranted.


jcfurey:
Pushed one more build fix (8d7f074...) to the Release workflow, extending the unsigned-release work to the signing step itself.

The release was still failing without signing secrets: on macOS, an empty CSC_LINK got read as a cert path (<projectDir> not a file); on Windows, the hardcoded azureSignOptions always forced Azure signing. Fixed both — unset the empty CSC_LINK, and inject the (public) Azure options from the workflow only when creds exist.

Both platforms should sign when creds exist and build unsigned when they don't.

jcfurey:
Pushed a regression-audit pass on top of the modernization: fixes for the LTC worklet URL (404 under file://), 4 Svelte 5 :has() CSS rules being pruned, and a broken npm start dev server — plus unit tests for sanitizeZipPath/sanitizeFileName/getMachineId (now 267 tests, svelte-check clean, build green). A few runtime paths (audio metadata, contenteditable editing, NDI/BMD, PDF) still need manual QA. Again, I'd be glad to split this into smaller PRs if that helps review. I realize this is a bit much.

vassbo:
Thanks, but there's a reason this has not been done as Svelte Transitions are broken in newer versions, so we would have to create a custom transition system first!

https://github.com/ChurchApps/FreeShow/pull/1512

vassbo:
It's very mixed yeah, too much for one PR. I suppose the linting is nice though.

jcfurey:
Sounds like a plan. I'll try to have everything together on my end/fork for starting a pr stack. This evening my time, I'll push the first pr. Thanks!

PR #3384, opened 2026-06-16
Security & dependency hardening: clear npm advisories + commit lockfile

jcfurey:
Hardens the dependency tree and fixes a few correctness findings, with no functional changes to the app.

Dependencies / advisories (kept after review — no runtime coupling):

@rollup/plugin-terser 0.4 → 1.0 (serialize-javascript RCE)
npm-run-all → npm-run-all2 (unmaintained package + its cross-spawn ReDoS)
tmp → ^0.2.7 + overrides for shell-quote ^1.8.4 / tmp ^0.2.7 (path traversal / injection)
add engines.node; commit package-lock.json and mark it generated so it collapses in diffs
Per review, the Electron 37→40, music-metadata 7→11, and fast-xml-parser 5.x bumps were reverted to their pins — they'll return as focused follow-ups with the breaking-change handling.

Source: small fixes surfaced by a static review (thumbnails lock-release, files.ts EXIF resolve, a shared IPC reply dispatcher, a REMOTE auth gate, minor robustness in servers/zip/csv/media). The contested audit changes flagged in review were reverted.

Tooling: manual CI workflow; repaired Playwright smoke test.

npm ci, build, and unit tests pass. (Line count is dominated by package-lock.json, collapsed via .gitattributes; reviewable source delta ≈14 files.)

PR #3385, opened 2026-06-16
Migrate to ESLint 9 flat config + replace new Function() math eval with a safe parser

jcfurey:
🔗 Stacked on #3384 (deps-and-security). Until that merges, the diff below also includes PR1's changes; GitHub collapses it to just this PR's delta once PR1 lands.

Security: replaces a new Function() arithmetic evaluator with a small, safe arithmetic parser in [src/frontend/utils/expression.ts](https://github.com/jcfurey/freeshow/blob/claude/compassionate-hawking-4w4tcd/src/frontend/utils/expression.ts) (+108), updating its call sites (apiHelper, timers, number inputs). Removes an eval-style code path.

Tooling: migrates ESLint 8 → 9 (flat config in [config/linting/eslint.config.mjs](https://github.com/jcfurey/freeshow/blob/claude/compassionate-hawking-4w4tcd/config/linting/eslint.config.mjs)), adds typescript-eslint 8 + eslint-plugin-svelte, removes the legacy .eslintrc files, and applies the resulting eslint --fix autofixes.

⚠️ Review note: this PR's own delta is 298 files, but ~293 are the mechanical eslint --fix pass (commit afedf4c). The substantive changes are two commits — safe-eval (b70680b) and the flat-config migration (0f10f95); review those first. Happy to split the autofix into its own PR if preferred.


vassbo:
Not sure if it's a good idea to stack it on the other PR that requires a lot of changes.

vassbo:
https://github.com/ChurchApps/FreeShow/commit/b70680bdb34b2b13f894ccdf64dbea9b0491c3f5 is overkill, it's fine to use `new Function` because it's a local app and not a web app.

jcfurey:
I'll drop the safe-parser. The only other thing here is the ESLint 8→9 flat-config move, which isn't really necessary on its own, so I'll close this. (Only note: ESLint 8 is end-of-life, so the flat-config is on my fork whenever you decide to move — no rush.)

PR #3386, opened 2026-06-16
Migrate to Svelte 5 + Vite 8 + TypeScript 5 (compatibility mode)

jcfurey:
🔗 Stacked on #3385 (ESLint 9 + safe-eval). Until it merges, the diff below also includes PR1+PR2's changes; GitHub collapses it once they land.

Upgrades the frontend stack to Svelte 5 / Vite 8 / TypeScript 5 in compatibility mode (no rune rewrite), preserving pre-migration type strictness so this PR stays a framework bump rather than a type overhaul (that's PR5).

Included:

The core Svelte 5 + Vite 8 + TS 5 migration
Close self-closing non-void HTML elements (Svelte 5 requirement) and enforce going forward
Zero-risk a11y fixes (alt text, icon-button labels, aria-selected) + acknowledge intentional autofocus; keyboard activation for focusable color swatches
ACCESSIBILITY.md (scope/roadmap for operator-UI keyboard a11y) and per-platform BUILDING.md
Dockerfile + .dockerignore for a reproducible Linux build; switch CI to npm ci
Verification: production build (npm run build) and the Playwright smoke test pass.

Follow-ups (kept out of this framework bump by design):

svelte-check reports ~76 migration type errors at this stage; they're resolved in the strict-mode PR (PR5). (svelte-check is not part of upstream CI.)
Two |global output-transition regressions from this migration are fixed in PR6.


jcfurey:
Re: the Svelte 5 transition "snap" (ref #1512) — flagging proactively, since it's the blocker cited when closing #3379.

Root cause. Svelte made transitions local by default at the 3→4 boundary (carried into 5): a local transition only plays when its direct parent toggles — not when a block above it ({#key}/{#each}/{#if}) is created/destroyed. FreeShow's output transitions live inside {#key show} / {#each currentOut}, so post-migration they snap to the next frame. (svelte-migrate adds |global automatically; the manual compat-mode migration didn't — that's what regressed.)

Fix. |global on the three output custom transitions — OutputTransition (text/media/PDF/effect output), the song-attribution text in Output.svelte, and ListView — plus a keying fix ({#key transitionId} instead of {#key show}) so the outgoing slide's text isn't orphaned during the global outro. svelte-check 0 / build clean.

In this stack the fix lands in #3389 (the regression-fixes PR at the tail), to keep this one a pure framework bump — so transitions on this branch in isolation still snap; they animate once #3389 is in. Happy to pull it forward into this PR if you'd prefer it self-contained.

Scope. This only restores the pre-Svelte-5 behavior — it deliberately does not touch the transition rework in #2169 (the identical-text flash / partial-fade that predates Svelte 5). FreeShow already has a custom transition system ([utils/transitions.ts](https://github.com/jcfurey/freeshow/blob/dev/utils/transitions.ts)); this just makes the existing transitions play again rather than adding a new one.

jcfurey:
Build note — import.meta in the companion (iife) builds. A minor regression the Vite 4→8 bump exposes: [src/server/remote/components/Auth.svelte](https://github.com/jcfurey/freeshow/blob/dev/src/server/remote/components/Auth.svelte) resolves the login logo via new URL("…/freeshow.webp", import.meta.url). The companion apps build as iife ([config/building/vite.config.servers.mjs](https://github.com/jcfurey/freeshow/blob/dev/config/building/vite.config.servers.mjs) → formats: ['iife']), and Vite 8 replaces import.meta with {} in iife output — so the URL's base is undefined and the remote login logo doesn't load.

It's existing code (unchanged in this PR) and only the remote companion hits it; the main app and the other three companions are fine. I can fix it by referencing the logo via its served public path instead of import.meta.url (works in iife) and push it into this PR — or leave it to you if you'd rather set the companion-app asset strategy. Preference?

vassbo:
Did you fix the Svelte transition issue? I would like to have only the related diff in a PR.

jcfurey:
```diff
# output/transitions/OutputTransition.svelte
-    <div class="transitioner" in:custom={inTransition || transition || {}} out:custom={outTransition || transition || {}} on:outrostart>
+    <div class="transitioner" in:custom|global={inTransition || transition || {}} out:custom|global={outTransition || transition || {}} on:outrostart>

# output/Output.svelte  (attribution line, mirror branch unchanged)
-            <p class="attributionString" transition:custom={transitions.text}>...
+            <p class="attributionString" transition:custom|global={transitions.text}>...

# slide/views/ListView.svelte
-                    <div class="center" transition:custom={transition}>
+                    <div class="center" transition:custom|global={transition}>

# output/layers/SlideContent.svelte
     let show = false
+    let transitionId = 0   // distinct identity per reveal so the |global outro doesn't orphan prev text
...
                 timeout = setTimeout(() => {
+                    transitionId++   // bump before reveal
                     show = true
...
-            {#key show}
+            {#key transitionId}
                 {#if show}
```

Yes — fixed. Two parts: (1) add `|global` to the output transitions (`OutputTransition` in/out, the attribution string in `Output.svelte`, and `ListView`), since Svelte 4 made transitions local-by-default and they otherwise snap inside the `{#key}`/`{#each}` blocks; and (2) key the slide reveal on a monotonic id instead of the boolean `show`, so the outgoing slide's text isn't left on screen during the `|global` outro.

Do note that this only has an effect on Svelte 4/5. On the current Svelte-3 `dev`, transitions are already global by default, so the diff is a no-op there — it can't really stand as its own PR against `dev`; it only does something once the migration is underneath it. The whole transition fix is ~10 lines across those 4 files (diff above), so it's easy to review on its own within the migration. Would you still like me to open up the dedicated pull request?

vassbo:
Okay, that makes sense. But I would still like it to not be stacked on another PR. Also have not checked if there's any other breaking changes.

jcfurey:
I'll get a new PR going for this sometime today. It'll likely be evening my time Central US.

vassbo:
No need to rush it, but it's probably good to get the other major PRs done and merged first?

jcfurey:
Sounds good. I'll be ready to rebase this and open up a fresh PR once you're happy with the state of #3384

PR #3387, opened 2026-06-16
Add a vitest unit-test suite for helpers and import converters

jcfurey:
🔗 Stacked on #3386 (Svelte 5 migration). Inherits PR3's deferred follow-ups (type errors → PR5, transitions → PR6).

Adds a vitest unit-test layer — almost entirely new files (+3153 / −3, plus vitest/jsdom dev-deps), no behavioral changes. 251 tests pass at this branch.

Coverage:

Pure helpers: expression, color, scripture, chord transpose, bytes, style, cropping, mover, clickable, languageData, array, time
Electron helpers, fonts, request
Import converters: ChordPro, CSV, txt, SongBeamer; XML converters via jsdom (xml2json, OpenSong, OpenLP, Zefania, ProPresenter, Quelea, OSIS/Beblia, VerseView, EasySlides, MediaShout, VideoPsalm, SoftProjector); Calendar (ICS) and Project import
Tests use dependency/collaborator mocking to isolate units. Run with vitest run --config config/testing/vitest.config.ts.

PR #3388, opened 2026-06-16
Enable TypeScript strict mode across frontend, electron, and server

jcfurey:
🔗 Stacked on #3387 (vitest suite). This PR clears the ~76 Svelte 5 migration type errors deferred from PR3/PR4 and completes strict mode.

Builds on the Svelte 5 migration to re-enable and complete TypeScript strict mode.

Resolve the remaining Svelte 5 migration type errors (svelte-check → 0)
Re-enable strict mode, then work through noImplicitAny in stages (type untyped params, cast index access, type variable declarations) until full strict passes
Enable full strict for the electron tsconfig; make the already-strict server tsconfig type-check clean
Mostly type annotations and safe casts (+410 / −415), no runtime behavior change.

Verification: svelte-check (0 errors), electron tsc, and server tsc all clean.


vassbo:
For what I can see it just adds `: any` explicitly in many cases to variables that already are detected as `any`.

PR #3389, opened 2026-06-16
Build/release robustness, Svelte 5 transition fixes, and regression fixes

jcfurey:
🔗 Stacked on #3388 (strict mode) — the final PR in the modernization stack.

Final polish on top of the modernization stack.

Svelte 5 transition regressions: restore |global output transitions (the "snap" blocker) and stop previous slide text lingering during |global transitions — fixes visible playback regressions introduced by the Svelte 5 migration (PR3)
Release: skip macOS notarization when Apple credentials aren't set; tolerate missing code-signing creds on macOS/Windows
Regression fixes + security tests surfaced during the modernization
Formatting: reformat with Prettier 3.8.4
Docs: Windows native-build gotchas + the unit-test layer
Verification: full svelte-check (0 errors) and vitest suite (267 tests) pass.


vassbo:
I would like to see the relevant diff.

jcfurey:
I'm just going to leave comments on the related work on #3386

PR #3390, opened 2026-06-16
Repair show/lyrics search scorer, add caching, and support exact-phrase queries

jcfurey:
Independent of the modernization stack — improves show/lyrics search. Builds on upstream's existing vitest setup (no new deps).

Repair the show/lyrics search scorer (rewrites search.ts; adds search.test.ts — 19 tests, all passing)
Add a lyric-search cache and stop search results flooding
Support quoted queries for strict exact-phrase matching
+285 / −106 across 7 files.


vassbo:
What's the improvement here?
And no need to have the Rebuild button.

jcfurey:
The main fix: search was returning almost every show for many queries — fuzzy similarity matched everything, so results flooded. Now a show is only included on a real match (exact word/phrase, content density, or a strong ≥0.7 typo match); similarity only re-ranks. The scorer also repairs a few concrete bugs: multi-word matching was disabled, removeShortWords was a no-op, content could outrank the title, and empty/punctuation-only queries matched everything. Quoting a query ("amazing grace") is an opt-in strict phrase match (per-search, no settings). 19 unit tests cover it.

And I've removed the Rebuild button — you're right, the cache auto-backfills on load, so it was redundant.

vassbo:
Nice, great!

PR #3391, opened 2026-06-16
Protect Bible text from accidental edits (scripture shows read-only by default)

jcfurey:
Adds an opt-out safeguard so downloaded/API Bible text can't be silently altered and re-projected as Scripture.

Verse text in Bible-sourced shows (reference.type === "scripture") is read-only by default — styling, layout, templates, and verse-number options stay editable; only the words lock.

Global setting "Protect Bible text from editing" (default on): special.protectScriptureText.
Per-show unlock in the edit-header "⋯" dropdown: show.unlockedScriptureText.
Gated paths (shared isScriptureTextLocked() helper, all respecting the setting + per-show unlock):

inline editbox (EditboxLines read-only render) + plain-text popup (TextEditor disabled);
format actions (capitalize/lowercase/uppercase/trim) + find & replace (format());
the formatText() rebuild (set_plain_text + transpose_show_* API, SlideEditor/TextEditor transpose);
the set_show API.
Structural ops (style/layout, merge/split, paste-slides, delete) stay editable — they don't alter the words. svelte-check 0 errors. 10 files, +56/−4.

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -999,6 +999,8 @@
     },
     "edit": {
         "text": "Text",
+        "unlock_scripture_text": "Unlock scripture text",
+        "unlock_scripture_text_tip": "Allow editing the protected Bible text in this show",
         "font": "Font",
         "family": "Font family",
         "font_size": "Font size",
@@ -1352,6 +1354,7 @@
         "next_item_on_last_slide": "Arrow left/right changes project item on first/last slide",
         "slide_number_keys": "Play slides with number keys",
         "auto_shortcut_first_letter": "Auto shortcut to first letter in text",
+        "protect_scripture_text": "Protect Bible text from editing",
         "startup_projects_list": "Show the projects list on startup",
         "auto_output": "Activate output screen on startup",
         "hide_cursor_in_output": "Hide cursor in output",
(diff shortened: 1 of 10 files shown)


vassbo:
I get why this makes sense, but I really don't think it's necessary. I don't think people will edit the scripture text willy-nilly, but rather fix spelling mistakes or break it up into parts, etc. We could lock the whole show by default but then people would have to unlock it to just change the styling.

jcfurey:
That's fair. One clarification in case it's useful: it only locks the verse text — styling, templates, layout, and splitting into separate slides all stay editable — and it's an opt-out toggle with a per-show unlock, so it's not locking the whole show. But you're right that fixing a typo means editing the text, which it would get in the way. I'll keep it on my fork for now for anyone who wants that guard rail.

vassbo:
> it only locks the verse text — styling, templates, layout, and splitting into separate slides all stay editable

Yeah, I got that.

jcfurey:
Ok. I was mostly concerned with the optics of it. I did leave the editable scripture as a global and a local scripture-show change lock change, should the issue warrant a revisit. I'll try to leave the relevant branch up on my fork if future issues crop up.

PR #3395, opened 2026-06-16
PCO Integration Improvements

schplay:
This PR contains several improvements/additions to the PCO integration. These updates are important for users coming from ProPresenter as they are hugely important for those that heavily utilize Planning Center Services across multiple service types as well as depend on the PCO "live" timing. My church falls into that category and the lack of these features has prevented us from even being able to seriously consider switching to FreeShow. I hope that these will also help other users be able to make the switch.

- New PCO Timer type allowing live PCO service time sync.
- Updated wording for countdown "overflow" to make it easier to understand.
- Added PusherJS to support PCO "live" events for new PCO Timer.
- New PCO Service type selection in the PCO config to allow customization of which service types/folders are auto-synced. This also fixes PCO import for all accounts with service type and folder structures that don't match the default assumption the app previously made which in some cases made the feature unusable.
- New ability when manually creating a Project to select any upcoming PCO service that would not have otherwise gotten synced and created.
- New manual refresh button when viewing a PCO Project to force re-syncing to get updates from online.

These were tested and are working well on Windows.


vassbo:
Awesome! This is great.
Can we instead of having it as a "Timer", we should just have it as a dynamic value.

vassbo:
Also we should use Socket.IO instead of PusherJS because that's already being used in the app.

schplay:
@vassbo concerning "Timer" vs a dynamic value I'm not 100% sure that I understand. Do you mean to have it be a "Variable" under the "Functions"? I honestly think having it as a "Timer" makes the most sense for users since that's essentially what it is; it's just controlled by an external service instead of logic within FreeShow. Furthermore if we were to change it to being a "Variable" then I don't currently see a way to display those on the Stage Display and we would need to find the best way to add that (presumably an entirely new "Item" type).

Unfortunately the existing Socket.io implementation isn't compatible with the Pusher events used by PCO so unless I'm mistaken there's not a way to replace one with the other.

vassbo:
Right click any textbox and you'll find "Dynamic values" - timers and variables are here as well.
Because it does not need to be controlled, we just need it as text and not as an interface.

schplay:
@vassbo I see, I wasn't aware of that feature. After looking at it I have some questions about how you suggest I would make that change. For instance, would it be a new dynamic value under the "Project" group or would it make sense to be a new "PCO" group? Also, if making this change from a "Timer" then how would we configure the early warning and overflow settings that are native to "Timers" but presumably don't exist for text?

vassbo:
I guess if it needs many values it can be it's own "PCO" section, if not it can be put in the "Project" section, you can choose.

Textboxes and dynamic values work well with "Conditions" where you can show/hide certain textboxes based on the value of a dynamic value. So it's up to the user to configure multiple textboxes with different colors here, you can just leave it as plain text.

vassbo:
I guess the condition would need an update to allow setting "is above"/"is below" if the dynamic value can be parsed into a number.

schplay:
I'm going to be honest, that sounds like a lot of work just to duplicate existing functionality that EVERY single person that uses this PCO timer will want. I think having it available as a Dynamic Value is a good idea but in the current implementation it already is since it's a timer. I also think new "is above" and "is below" modes for the conditional logic are good ideas but I think requiring every user to build their own logic there just to get the existing timer functionality is bad UX.

vassbo:
Then we should just make a default PCO timer stage display (and/or template) that is created when the user connects with Planning Center, that can also be used as an example.

schplay:
I don't see any concept of a template for the display layout or display items so presumably your preference would be to have the application programmatically create a new PCO stage display for the user when they initially connect the PCO integration?

vassbo:
Correct.

schplay:
@vassbo the PCO timer has options that change how the countdown is calculated: "End Item On Time", "Full Item Length", and "End Service On Time". If rebuilding the Timer using Dynamic Values and Conditions, how would you suggest we handle those options? This is something that may need to be changed on a regular basis by operators and the more we flesh out this solution of building logic blocks the more of a nightmare that sounds for regular volunteers that would need to use this.

vassbo:
I think we can just have seperate dynamic values for each of them all live at once.

schplay:
I'm just thinking about the perspective of a regular user/volunteer. If they're told to update a timer mid-service to display "Full Item Length" for a specific moment/time during service and then change it back to "End Item On Time" afterwards, having that be a simple dropdown item on an actual Timer that is named "PCO Timer" is simple and easy to understand/remember. With what you're proposing, instead they have to edit a stage display, be trained on and understand Dynamic Values, and then manually change out those values while being careful not to make any typos or other mistakes that affect the stage display. That's going to be difficult for some people to grasp and remember let alone carry out quickly and without mistakes. 

I just don't understand how this is a better solution or what we gain by changing this timer from being an actual Timer. There is already a concept of different types of timers in the application and there's a reason that they were built in the way that they are instead of all being built like you're suggesting we do for this one.

vassbo:
I think if it's a Timer, or a Dynamic value I would create seperate Stage displays for the different timers and use actions to automate changing between those so that should not matter in the use. Only difference is in the setup.

vassbo:
In my mind since those timers "already exists", it makes more sense that they are a dynamic value rather than you having to "create" a new timer that already is running somewhere.

schplay:
> I think if it's a Timer, or a Dynamic value I would create seperate Stage displays for the different timers and use actions to automate changing between those so that should not matter in the use. Only difference is in the setup.

I think that could work for some people in some situations but it doesn't cover all scenarios and carries it's own drawbacks. For example, that would require editing multiple stage displays anytime you need to update one.

> In my mind since those timers "already exists", it makes more sense that they are a dynamic value rather than you having to "create" a new timer that already is running somewhere.

This is where we fundamentally differ. While the instructions for the timer (IE the amount of time being counted down) are coming from an external source, the application is still handling the actual counting down of time. In that sense it's identical to the other "Timer"s. The difference is that instead of manually entering an amount of time to count down, you select a PCO Service which will dynamically update the amount of time that the application should count down as you progress through a service. The way that the user thinks of it, interacts with it, etc is just like the other "Timer"s and is also how other presentation software treats it. While I think the Dynamic Value, Condition, and Action features in FreeShow are great I also think that trying to use them instead of a "Timer" to implement a timer doesn't make much sense.

vassbo:
With a bit more setup you can actually use just one stage display, with for example three different textbox timer items, and create toggles with Variables to turn them on/off via Conditions. So it's not an issue if the person setting it up knows what he is doing.

But okay, about Timers it sounds like you would need to update the timer for each new PCO service? That sounds unnecessary.

With dynamic values it would make sense they would just change automatically based on the opened project so you don't need to change it.

vassbo:
Regardless if it's Timers or Dynamic values I think because we only pull data and don't push anything, we don't need to open a Pusher Socket connection, we can simplify the code a lot by just sending HTTP requests. Here's how I would do the connection (currently to dynamic values): https://github.com/ChurchApps/FreeShow/tree/pco-live

schplay:
> With dynamic values it would make sense they would just change automatically based on the opened project so you don't need to change it.

The only problem with this is if you need to open and edit a Project while another project is ongoing: doing so would mess up the live clock for those depending on it while you are making your edits. This can easily happen if you have multiple services in a day and they aren't built/edited ahead of time.

> Regardless if it's Timers or Dynamic values I think because we only pull data and don't push anything, we don't need to open a Pusher Socket connection, we can simplify the code a lot by just sending HTTP requests. Here's how I would do the connection (currently to dynamic values): https://github.com/ChurchApps/FreeShow/tree/pco-live

The Pusher Socket is so that we can receive real-time events for the service clock changes. If we were to use HTTPS requests we would need to constantly poll the server which would add delays or would hit the API rate limits. Using the Pusher Socket is the correct way of handling this use-case.

vassbo:
> if you need to open and edit a Project while another project is ongoing

When have that been an issue? I have never ever done that, and I can't see why you would do that. - Maybe we can set the Timer to automatically use the opened project then, so people won't need to constantly go and update it?!

> If we were to use HTTPS requests we would need to constantly poll the server which would add delays or would hit the API rate limits.

Gotcha.

src/frontend/components/main/popups/Timer.svelte
```diff
@@ -36,9 +36,26 @@
     const timerTypes: any = [
         { id: "counter", name: translateText("timer.from_to"), translate: true, icon: "timer" },
         { id: "clock", name: translateText("timer.to_time"), translate: true, icon: "clock" },
-        { id: "event", name: translateText("timer.to_event"), translate: true, icon: "calendar" }
+        { id: "event", name: translateText("timer.to_event"), translate: true, icon: "calendar" },
+        { id: "pco_live", name: "PCO Live Countdown", icon: "calendar" }
```
vassbo: No need to show if not logged in to Planning Center. And we should have a unique icon.

src/frontend/components/main/popups/Timer.svelte
```diff
@@ -290,6 +312,30 @@
...
+                    <MaterialDropdown
+                        label="Service plan"
+                        style="width: 100%;"
+                        options={pcoPlanOptions}
+                        value={(timer.pcoPlanId && timer.pcoServiceTypeId) ? timer.pcoPlanId + "|" + timer.pcoServiceTypeId : ""}
+                        on:change={updatePcoPlan}
+                    />
+                {/if}
+                <MaterialDropdown
+                    label="Countdown type"
```
vassbo: Needs localization.

src/frontend/components/main/popups/Timer.svelte
```diff
@@ -314,7 +360,7 @@
     {/if}
 
     <InputRow arrow={timer.warn} bind:open={warningMenuOpened}>
-        <MaterialToggleSwitch label="timer.warn_early" style="width: 100%;" checked={timer.warn} defaultValue={false} on:change={(e) => updateTimerValue("warn", e.detail)} />
+        <MaterialToggleSwitch label="timer.warn_early" style="width: 100%;" checked={timer.warn} defaultValue={true} on:change={(e) => updateTimerValue("warn", e.detail)} />
```
vassbo: Keep at false

src/frontend/components/main/popups/Timer.svelte
```diff
@@ -324,7 +370,7 @@
     </InputRow>
 
     <InputRow arrow={timer.overflow} bind:open={overflowMenuOpened}>
-        <MaterialToggleSwitch label="timer.overflow" style="width: 100%;" checked={timer.overflow} defaultValue={false} on:change={(e) => updateTimerValue("overflow", e.detail)} />
+        <MaterialToggleSwitch label="timer.overflow" style="width: 100%;" checked={timer.overflow} defaultValue={true} on:change={(e) => updateTimerValue("overflow", e.detail)} />
```
vassbo: Keep at false

src/frontend/components/settings/tabs/Connection.svelte
```diff
@@ -284,6 +296,42 @@
...
+    {:else if folderTree.length === 0}
+        <Tip type="warning" value="No folders found in your Planning Center account." top={0} />
+    {:else}
+        <InputRow>
+            <MaterialButton
+                style="flex: 1;"
+                on:click={() => updateProvider("planningcenter", "selectedFolderIds", [])}
+                title="Clear selection to sync all services"
+            >
+                Clear selection (sync all)
```
vassbo: Needs localization.

src/frontend/components/show/PcoServiceTreeNodes.svelte
```diff
@@ -0,0 +1,117 @@
...
+            <button class="expand-btn" on:click={() => dispatch("toggle", node.id)} title={expanded.has(node.id) ? "Collapse" : "Expand"}>
+                <Icon id={expanded.has(node.id) ? "arrow_down" : "arrow_right"} size={0.75} white />
+            </button>
+            <Icon id="list" size={0.8} white />
+            <span class="label">{node.name}</span>
+        </div>
+        {#if expanded.has(node.id) && node.children.length}
+            <svelte:self nodes={node.children} {expanded} depth={depth + 1} on:toggle={forwardToggle} on:select={forwardSelect} />
+        {:else if expanded.has(node.id)}
+            <div class="no-plans" style="padding-inline-start: {(depth + 1) * 16 + 28}px">No upcoming services</div>
```
vassbo: Needs localization.

src/frontend/App.svelte
```diff
@@ -32,6 +33,7 @@
     // countdown timer tick
     $: if ($activeTimers.length) startTimer()
     $: if (Object.keys($timers).length) checkTimers()
+    $: if ($timers) initPcoLiveSync($timers)
```
vassbo: Double listener, use checkTimers

src/types/Show.ts
```diff
@@ -191,6 +191,10 @@ export interface Timer {
     endDynamic?: string
     event?: string
     time?: string
+    // PCO Live countdown fields
+    pcoServiceTypeId?: string
+    pcoPlanId?: string
+    pcoCountdownType?: "full_length" | "end_on_time" | "end_service"
```
vassbo: We should have one pco object instead of three.

src/electron/contentProviders/planningCenter/live.ts
```diff
@@ -0,0 +1,133 @@
...
+    const liveChannel: string | null = liveList[0].attributes?.live_channel ?? null
+
+    // Live resource is a singleton — endpoint is live/current_item_time, NOT live/{id}/current_item_time
+    const itemTimeList = await pcoRequest({ scope: "services", endpoint: `service_types/${serviceTypeId}/plans/${planId}/live/current_item_time` })
+
+    if (!itemTimeList?.length) return { liveId, liveChannel, orgId, liveStartAt: null, liveEndAt: null, length: null, isPreService: false, serviceStartAt, serviceEndAt }
+
+    const itemTime = itemTimeList[0]
+    if (!itemTime?.attributes) return { liveId, liveChannel, orgId, liveStartAt: null, liveEndAt: null, length: null, isPreService: false, serviceStartAt, serviceEndAt }
+    if (itemTime.attributes.exclude) return { liveId, liveChannel, orgId, liveStartAt: null, liveEndAt: null, length: null, isPreService: false, serviceStartAt, serviceEndAt }
```
vassbo: Three identical returns.

vassbo:
Maybe the PCO Live Timer should have been a seperate PR to the other changes?
I don't see why we need the Service Picker as that is done automatically.

schplay:
> When have that been an issue? I have never ever done that, and I can't see why you would do that. - Maybe we can set the Timer to automatically use the opened project then, so people won't need to constantly go and update it?!

I think this is a source of problems: everyone has different situations and workflows and just because it's not your own doesn't make it invalid. For example, the scenario I described is one that I experience every week. Our volunteer that runs our presentation software (currently ProPresenter) arrives and turns on the stage display while the Production Manager begins the service clock which includes pre-service items and timings. That timer is used by the Worship and Production ministries immediately while the volunteer begins to build out the content for the services that day. They usually start with the first service and then move to the second service. This takes some time and the operator may need to jump back and forth a handful of times for various reasons during which the stage display needs to be showing the PCO timer for the current service.

> I don't see why we need the Service Picker as that is done automatically.

This again is the same issue where your experience is not all-encompassing. The existing PCO integration is broken because it does not account for folders when finding Services in PCO. For people that have a large variety of events/services that are managed in PCO this is a problem. For example, with the existing integration FreeShow was unable to find ANY of our current services and instead only shows a handful of archived Services from over a year ago. Adding awareness of how PCO allows folder organization was the first step in fixing this. If you have multiple spaces/rooms/etc for different Services/events you don't want every computer to load every Service. To account for this I added the ability to customize which PCO Service folders are used for the auto-sync feature. Then to account for any time that you want to manually sync a specific Service I added the ability to do so. For example, if you have a space for Youth that computer can be set to sync only Youth Services while your main sanctuary/auditorium can be set to sync only Sunday Services. Then if you have a one-off Youth event in the main sanctuary you can just manually sync that one specific service without changing your normal sync settings. This should cover the majority of users/scenarios.

vassbo:
Gotcha.

src/frontend/components/show/ProjectList.svelte
```diff
@@ -102,11 +104,22 @@
 
     function open(e: any, id: string) {
         if (e.detail.target.closest(".edit") || e.detail.target.querySelector(".edit") || editActive) return
+        if (e.detail.target.closest(".pco-refresh")) return
         if (e.detail.ctrl) return
 
         openProject(id, !e.detail.alt)
     }
 
+    function getProjectFolderId(projectId: string): string {
```
vassbo: Name should contain PCO, or explain better what it does.

src/frontend/components/show/PcoServicePicker.svelte
```diff
@@ -0,0 +1,128 @@
...
+        <button class="closeBtn" on:click={() => dispatch("close")}>
+            <Icon id="close" size={0.9} white />
+        </button>
+    </div>
+
+    <div class="pickerBody">
+        {#if loading}
+            <p class="status">Loading…</p>
+        {:else if fetchError || !nodes.length}
+            <p class="status empty">No upcoming services found.</p>
```
vassbo: Needs localization.

vassbo:
I just needed some explanations. I will open a review.

src/frontend/components/show/Projects.svelte
```diff
@@ -496,9 +519,17 @@
                             <Icon id="folder" size={0.7} white />
                         </div>
                     </MaterialButton>
+
+                    <MaterialButton variant="outlined" icon="calendar" title="Planning Center" on:click={openPcoPicker} white>
+                        Planning Center
```
vassbo: No need to show this if not logged in to Planning Center.

vassbo (changes requested):

src/frontend/components/drawer/timers/Timers.svelte
```diff
@@ -116,7 +117,11 @@
             {#if i === 0 && list.length}
                 <h5><T id="remote.project" /></h5>
             {:else if title && i > 0}
-                <h5><T id={timerTypeNames[title]} /></h5>
+                {#if title === "pco_live"}
+                    <h5>PCO Live</h5>
```
vassbo: This is unnecessary because it's already in the timerTypeNames.

vassbo (changes requested):

vassbo:
I still think it would be the best solution for the general user to have it automatically change the timer based on the opened project, that can be the first option in the dropdown, or you can choose a specific project/service if you wish to do that. - Because I don't think the majority of users change back and forth between projects - that's what I meant when I requested an explanation.

schplay:
I think that's fine as the default behavior while keeping the ability to manually specify. I'll work on the rest of these change requests as well.

vassbo:
Great! I rewrote all the UI so it matches the rest of the app. It's good now.

vassbo:
@schplay https://github.com/ChurchApps/FreeShow/releases/tag/v1.6.4-beta.1

PR #3396, opened 2026-06-17
Migrate to Svelte 5 + Vite 8 + TypeScript 5 (compatibility mode)

jcfurey:
Standalone migration of the frontend + companion apps to Svelte 5 / Vite 8 / TypeScript 5, compatibility mode (no rune rewrite). Rebased onto current dev (now includes #3384). Pre-migration type strictness is preserved, so this stays a framework bump, not a type overhaul.

The reviewable surface is small — the 51-file count is the regenerated package-lock.json (GitHub collapses it) plus ~50 one-line changes:

- Core: main.ts new App() → mount(App, …) (+ each companion main.ts); svelte.config.mjs / vite.config.mjs / [config/building/vite.config.servers.mjs](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/config/building/vite.config.servers.mjs) switch svelte-preprocess → vitePreprocess and filter Svelte 5's renamed a11y_* warning codes (legacy a11y- branch dropped per review, in both vite configs); vite.config.mjs assetFileNames reads assetInfo.names[] (Vite 8 / Rollup 4); package.json framework bumps + drop the now-incompatible Svelte-3 tooling; [src/frontend/tsconfig.json](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/src/frontend/tsconfig.json) via @tsconfig/svelte v5 (strict umbrella kept disabled — compat mode).
- ~28 import type conversions — Vite 8's rolldown rejects type-as-value imports.
- <script type="ts"> → lang="ts" in Top.svelte.

Transition fix (the #1512 "snap"), folded in so the PR is self-contained:

- |global on the output transitions (OutputTransition, the attribution text in Output.svelte, ListView).
- SlideContent keys the reveal on a monotonic transitionId instead of the boolean show, so the outgoing slide's text isn't orphaned during the |global outro.

Real fixes surfaced by the stricter toolchain:

- [src/electron/data/bonjour.ts](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/src/electron/data/bonjour.ts) — Service imported as a value + InstanceType<typeof Service> (it's a class).
- Connection.svelte — local connections renamed to connectionCount (it shadowed the store import).
- [src/server/remote/components/Auth.svelte](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/src/server/remote/components/Auth.svelte) — the login logo used new URL(…, import.meta.url), which throws "Invalid URL" at runtime in the Vite 8 iife companion build; now a plain asset import (Vite inlines the 378-byte webp as a data URI). *.webp ambient type added.

Tooling: committed package-lock.json; vitest 2 → 4 (vitest 2 dragged in vite 5 / esbuild 0.21, blocking npm ci on vite 8); dropped eslint-plugin-svelte3 + its lint:svelte step (Svelte-3-only, blocks install on Svelte 5) — .svelte stays covered by svelte-check / prettier / stylelint.

Compatibility-mode caveat: svelte-check reports ~71 type errors at this stage (the @tsconfig/svelte v5 strict flags, deferred by design). The production build is unaffected, and svelte-check isn't in upstream CI.

Verification: npm ci ✓ · npm run build ✓ (frontend + companion servers + electron) · unit tests pass (vitest 4) · built app launches and renders.


config/linting/eslint.svelte.js
```diff
@@ -1,45 +0,0 @@
-module.exports = {
```
vassbo, line 1: Do we not need this file?

src/frontend/components/draw/DrawSettings.svelte
```diff
@@ -126,7 +126,7 @@
             <T id="clear.drawing" />
         </MaterialButton>
 
-        <div class="divider" />
+        <div class="divider"></div>
```
vassbo: We can keep  `/>` still in all the places!

config/building/vite.config.servers.mjs
```diff
@@ -128,17 +127,13 @@ function copyServerFilesPlugin(serverId) {
...
-      }),
+      preprocess: vitePreprocess(),
       compilerOptions: {
         dev: !production,
       },
       onwarn: (warning, handler) => {
-        // disable A11y warnings
-        if (warning.code.startsWith('a11y-')) return
+        // disable A11y warnings (Svelte 5 renamed codes from a11y-* to a11y_*)
+        if (warning.code.startsWith('a11y-') || warning.code.startsWith('a11y_')) return
```
vassbo: Okay, but we don't need the old check then.

src/server/remote/components/Auth.svelte
```diff
@@ -3,8 +3,10 @@
     import { translate } from "../util/helpers"
     import { send } from "../util/socket"
     import { _get, _update, dictionary, password } from "../util/stores"
-
-    const freeshowLogo = new URL("../../../../public/import-logos/freeshow.webp", import.meta.url).href
+    // import the asset directly so Vite resolves it (small → inlined data URI); avoids
+    // `import.meta.url`, which Vite 8's iife companion build replaces with {} and would
+    // throw "Invalid URL" at runtime.
+    import freeshowLogo from "../../../../public/import-logos/freeshow.webp"
```
vassbo, line 7: Check this.

vassbo:
Check my reviews, and we probably don't need the new files.

vassbo (changes requested):

src/frontend/components/draw/DrawSettings.svelte
jcfurey: 😔 I'll work on reverting those if you'd like.

vassbo: This is a lint change, right?

jcfurey: Yes it was an ESLint autofix (the svelte/html-self-closing rule). Reverted all of them (kept /> everywhere) and dropped the rule, so the migration doesn't touch them anymore.

config/linting/eslint.svelte.js
jcfurey: It's the Svelte-3 ESLint config (plugins: ["svelte3"] / processor: "svelte3/svelte3"), which pulls in eslint-plugin-svelte3. That's Svelte-3-only and blocks npm install on Svelte 5 (peer conflict). .svelte stays covered by svelte-check + prettier + stylelint; a modern eslint-plugin-svelte flat config belongs with the separate ESLint 9 work.

config/building/vite.config.servers.mjs
jcfurey: Done. Dropped the legacy a11y- check, kept only a11y_.

src/server/remote/components/Auth.svelte
jcfurey: Verified: the companion apps build as iife, where Vite 8 blanks import.meta to {}, so the old new URL(logo, import.meta.url) throws "Invalid URL" at runtime. The asset import inlines the 378-byte webp as a data URI (no import.meta) and the iife server build passes. If you'd rather switch approaches, I'm happy to do that as well.

vassbo: I think both works, but I got `Cannot find module '../../../../public/import-logos/freeshow.webp' or its corresponding type declarations.` warning. Do you think Vite 8 resolves that?

src/frontend/main.ts
```diff
@@ -17,6 +17,7 @@ Sentry.init({
     }
 })
 
-const app = new App({ target: document.body })
+// Svelte 5: components are instantiated with mount() instead of `new App()`
```
vassbo: Redundant ^

src/frontend/components/output/layers/SlideContent.svelte
```diff
@@ -48,6 +48,11 @@
     let currentItems: Item[] = []
     let current: any = {}
     let show = false
+    // Monotonic id for the {#key} below. Keying on the boolean `show` makes the outgoing slide
+    // (key=true) and the incoming slide (key=true again) share an identity, so Svelte can't keep
+    // both alive during a |global transition — the previous slide's text gets left on screen.
+    // A unique id per reveal gives each transition its own identity so the outro cleans up.
```
vassbo: Maybe shorten the comment

vite.config.mjs
```diff
@@ -37,10 +32,12 @@ export default defineConfig({
               rollupOptions: {
                   output: {
                       assetFileNames: (assetInfo) => {
-                          if (assetInfo.name.endsWith(".css")) {
+                          // Vite 8 / Rollup 4: prefer names[] (the singular .name is deprecated)
+                          const name = assetInfo.names?.[0] ?? assetInfo.name ?? ""
```
vassbo: Old check might be redundant?

vassbo (changes requested):

src/server/remote/components/Auth.svelte
jcfurey: Two layers here:

Runtime — yes, Vite 8 resolves it. The companion build inlines the 378-byte webp as a base64 data URI directly into the iife bundle (no network request, no import.meta) — which is exactly why I moved off new URL(…, import.meta.url), since that's what threw "Invalid URL" in the iife build.

Types — the Cannot find module '*.webp' warning is separate, and it's already handled in-repo. Asset imports need an ambient module declaration, so I added one in [src/server/remote/global.d.ts](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/src/server/remote/global.d.ts), same pattern as the existing *?url / *.svelte declarations in [src/frontend/global.d.ts](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/src/frontend/global.d.ts):

declare module "*.webp" {
    const url: string
    export default url
}
I checked it's actually in scope, not just present:

npm run test:svelte (svelte-check) is clean on Auth.svelte — 0 "Cannot find module" across the whole run (the 71 it reports are all the deferred @tsconfig/svelte v5 strict flags, none asset-related).
tsc --showConfig -p src/server/tsconfig.json lists [./remote/global.d.ts](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/remote/global.d.ts) in its files (so does [config/typescript/tsconfig.server.json](https://github.com/jcfurey/freeshow/blob/feat/svelte5-vite8-migration/config/typescript/tsconfig.server.json)).
So the squiggle is a stale TS/Svelte language server that hasn't picked up the new .d.ts — Restart TS Server (or reload the window) clears it. If you'd rather not hand-roll the declaration, I can swap it for /// <reference types="vite/client" /> (Vite ships *.webp in its client types), but I kept it consistent with the existing *?url pattern.

vassbo: Thanks. I guess this will work then, but we can shorten the comment.

vassbo:
Let's hope this works fine. I will test a bit on a seperate branch first.

vassbo:
@jcfurey Btw. I did look at this but the transitions did not work properly. They were not smooth.

jcfurey:
Alrighty I'll take a look at it this week

jcfurey:
I think the transitions are rough because the branch is missing the transition fixes, not because of the migration.

`svelte5` is at 1.6.2-beta.2, `dev` is at 1.6.5-beta.3, and there's two commits thatmatter:

- #3402 added the `isDifferentSlide` check before setting `transitioningBetween`
  in `SlideContent.svelte`. Without it the between-transition runs even when the
  slide hasn't changed.
- #3517 added the "prevent stacking of the same item on update" early return in
  `SlideItemTransition.startTransition()`. Without it each reactive re-run mounts
  another `OutputTransition` with its own in/out, so the animations layer on top
  of each other.

Both landed after e12355a, where this branch was cut.

Would you like to rebase the svelte5 migration onto current `dev` so you can test it against 1.6.5-beta.3? I'll try that out on my fork. I'll play with it this week to test live use.

vassbo:
That makes sense, thanks.

> Would you like to rebase the svelte5 migration onto current dev so you can test it against 1.6.5-beta.3?

Sure. :)

Issue #3318, opened 2026-05-31
Suggestion Chord Diagramm implementation for "Slides"


BCK-Church:
In our church we digitilized our song chords in order to show them via monitors to our players. In general, FreeShow enables you to integrate pictures into presentations, but for us it would be really nice to just implement jpgs into a presentation or slides. For this implementation, it would also be necessary that an NDI source can direct to these chord diagrams so that when a slide with lyrics is presented, at the same moment the music players get the corresponding chord diagram on their monitors. We would really appreciate such a feature, or if we missed that this feature is already implemented, a short tutorial on where to find it would be fine.

We think this feature is also useful for other users because you could integrate pictures, side notes, or stage monitor files via this feature as well, so it has a high value for the program.

vassbo:
You can already do that. Create multiple outputs, and have the images items on a slide be sent to one output, and the text to another for example. You can right click any item on a slide and choose "Specific outputs".

BCK-Church:
If I understand correctly, you're able to integrate pictures behind every slide of a presentation by doing that? 

So that every slide inside a presentation gets aligned with a certain picture, even if you have different pictures for different slides? If you don't understand what I'm referring to, I meant it like you could do it in ProPresenter 7, where you can assign an Accordeogram to every slide.

BCK-Church:
@vassbo Aditionally we have around 1000 songs/presentations. Each presentation contains multiple slides. We would like to assign exactly one image per presentation containing the sheet music (JPG/PNG) so that this image is sent as an NDI source to a monitor for the musicians, synchronized with the presentation. 
Allow a single image file (sheet music) to be attached at the presentation level so it does not need to be copied onto every slide.
When switching slides during a song, only the lyric/text slides should change while the presentation-level sheet-music image is continuously sent as an NDI source to the musicians' monitors.
The sheet-music image should be routable as its own NDI output (separate from the main program/lyrics output).

vassbo:
> If I understand correctly, you're able to integrate pictures behind every slide of a presentation by doing that?

Yes, pretty much. You can have different items on the same slide that goes to different outputs.

You can also use actions to lock/unlock certain outputs, so you can change the lyrics on one output while the other one stays on a specific slide for example.

So everything you want to do is possible with a bit of tinkering. :)

BCK-Church:
Hmm, yes, I think I now understand how to make this work for us. The last thing I wanted to ask about is the roughly 1000 songs we have.

[...]

PR #3430, opened 2026-06-21
Add chord image (Akkordogramm) feature with NDI output selection

BCK-Church:
## What does this feature do?

Adds the ability to assign a chordogram image (JPG/PNG) to each song, 
which is automatically sent to a user-selectable output as soon as a slide 
for that song is activated.

## Use Case

Churches/bands that want to display chord diagrams for musicians on a separate monitor (e.g., via NDI) 
without them being visible on the main projector.

## Changes

- New `chordImage` and `chordOutputId` fields in Show Metadata
- UI in the Metadata tab for selecting the image and target output
- Logic in `showActions.ts` to send the image to the selected output when the slide changes

### DIFF
--- a/src/frontend/components/helpers/showActions.ts
+++ b/src/frontend/components/helpers/showActions.ts
@@ -307,21 +307,18 @@ export function updateOut(showId: string, index: number, layout: LayoutRef[], ex
                 }
             })
         }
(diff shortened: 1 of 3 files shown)


vassbo:
Thanks, but this is not the best way of doing this.

https://github.com/ChurchApps/FreeShow/issues/3318

PR #3443, opened 2026-06-22
Add flashing item background

eurdavi:
This PR adds a flashing "stage message" feature for stage text items, as discussed in the issue (ProPresenter-style stage messenger).

## What it does
- A stage text item can be marked to flash, with a configurable flash color and number of flashes.
- The background of the item pulses N times and then stops, leaving the text fully readable (the text stays visible the whole time).
- The flash triggers when a text variable referenced by the item is enabled (i.e. when the message is sent to the stage), and stops if that variable is disabled — so the flash is tied to the message being active.

## How it works
- Added optional fields to `StageItem`: `flash`, `flashColor`, `flashCount`.
- Added the controls (toggle, color, count) to the stage text item options in `BoxStyle.svelte`.
- Added a shared `FlashBackground` component (duplicated in the frontend and server bundles, since they don't share Svelte modules) that runs the pulse animation via the Web Animations API for a deterministic restart.
- Both Stagebox renderers (editor preview and the real stage output / localhost server) detect the referenced variables' enabled transitions and trigger/stop the flash. A module-level cache is used so the trigger survives component re-creation, and the initial mount does not fire.

## Notes / open to feedback
I made some pragmatic choices for my own use and I'm happy to adjust them to match the project's conventions — for example the label currently reuses an existing translation key, and I'd appreciate guidance on translations and on whether you'd prefer a different design (e.g. a dedicated item type instead of reusing the text item). Let me know what you'd like changed.

I've tested it working in the editor preview, the output window, and the stage web client. Video demo: 

https://github.com/user-attachments/assets/2e13f2e4-6562-4eba-8891-10758a19f2c2


vassbo:
Cool.

Please add it under the "Special" section, and only make the Color/Count show up when Flash is enabled.
Also trim down the long comments.
And I think we can do it without the context="module" scripts.
And instead of `flash`, `flashColor` and `flashCount`, can we do `flash: {enabled: true, ...}`

eurdavi:
> Cool.
> 
> Please add it under the "Special" section, and only make the Color/Count show up when Flash is enabled. Also trim down the long comments. And I think we can do it without the context="module" scripts. And instead of `flash`, `flashColor` and `flashCount`, can we do `flash: {enabled: true, ...}`

Changes made; thank you for the corrections. It is a pleasure to collaborate on this project.

eurdavi:
That's great that it worked out. Is there any estimate on when this addition might arrive in an update?

vassbo:
I simplified it at lot, and made it also work with Conditions and not just variables! Tomorrow!

Issue #3485, opened 2026-07-06
Import of ProSong .pro6 files does not work as expected


Toaster996:
**Describe the bug**
We are using [ProSong](https://prosong.ch/) to manage our song library and are considering replacing ProPresenter in some places with Freeshow. One mayor bug thats preventing us is that the Import of Songs exported from ProSong does not work expected. 
Since FreeShow claims to be able to import ProPresenter files, I classify this as a bug.

**Reproducing**
- import Do It Again example provided
- supposed to look anything like this
<img width="314" height="197" alt="Image" src="https://github.com/user-attachments/assets/93672837-f358-41c0-ba50-fd769cf36f5a" />

- however some Texts and all formating is missing

[Do It Again.pro6.zip](https://github.com/user-attachments/files/29701317/Do.It.Again.pro6.zip)

Toaster996:
<img width="913" height="520" alt="Image" src="https://github.com/user-attachments/assets/a7fbbe87-1d1b-455c-9b9c-099f695c5c27" />
This is how the import looks like in FreeShow
Only the first line is imported, the rest is somehow missing

vassbo:
I tried importing the .pro6 file you attached after merging the PR and it did not import.

vassbo:
@Toaster996 ??

Toaster996:
can you provide further details on what did not work?
for me, it worked on the current dev branch

<img width="595" height="312" alt="Image" src="https://github.com/user-attachments/assets/87e06bcb-625c-4996-af5e-363f24411dd9" />

On the first slide, i applied a theme, the others now have 4 text boxes as expected.

vassbo:
Thanks, I suppose it's a different issue as it did not even show up in the list for me, but I'll check it out.

vassbo:
I got it importing but it does not look so great because there are textboxes overlapping:
<img width="907" height="651" alt="Image" src="https://github.com/user-attachments/assets/2161d6fd-ea5d-4b07-8c03-9a082897cc37" />

vassbo:
But I suppose that's not an issue because it can be styles with templates.

Toaster996:
Yeah, not sure how to import the style, but with the fix we at least have the text boxes. Is the import supposed to also import the style?

vassbo:
It should get the item position in the slide when it exists. But not the text style.

Toaster996:
Ok, it does not yet do that - but also did not do that before. I check if this can be further improved

[...]

PR #3487, opened 2026-07-07
Fix import of pro6 files

Toaster996:
Fix for #3485 
- creates a list of items as expected
- refactors the `propresenter.ts` to improve readability


vassbo:
Nice, looks good! Thank you. :)

PR #3489, opened 2026-07-07
test: add unit tests for helpers, utils, and converters

Toaster996:
Since I think the test coverage for this project is a bit sad, I was motivated to add some tests....

Adds 9 test files covering ~160 new tests across the frontend and electron utility layers. Suite grows from 37 → 199 tests, all passing.

Frontend helpers (src/frontend/components/helpers/):
- array.test.ts     — sort/move/dedup/select helpers, defensive guards
- bytes.test.ts     — formatBytes across all size tiers
- color.test.ts     — hex/rgb/hsl round-trip, gradient parsing, contrast
- style.test.ts     — getStyles / getFilters / removeText
- time.test.ts      — secondsToTime, joinTime, timeAgo, date formatting

Frontend utils (src/frontend/utils/):
- chordTranspose.test.ts     — transposeText with slash chords, unicode,
                                sharp/flat preference, section labels
- common.test.ts             — throttle (leading + trailing edge),
                                hasNewerUpdate (superseding semantics), wait
- search.tokenize.test.ts    — tokenize + isRefinement

Electron helpers (src/electron/utils/):
- helpers.test.ts   — clone, checkIfMatching, wait, waitUntilValueIsDefined

Follows the mocking pattern established by the existing syncLedger and search tests — stub svelte stores / IPC modules; keep pure collaborators real. All files prettier-clean.

### DIFF
--- a/src/electron/utils/helpers.test.ts
+++ b/src/electron/utils/helpers.test.ts
@@ -1,0 +1,66 @@
+import { describe, it, expect } from "vitest"
+import { checkIfMatching, clone, wait, waitUntilValueIsDefined } from "./helpers"
+
+describe("clone (electron helper)", () => {
+    it("returns a deep copy — the original is untouched after mutations", () => {
+        const src = { a: { b: [1, 2, 3] } }
+        const copy = clone(src)
+        copy.a.b.push(99)
+        expect(src.a.b).toEqual([1, 2, 3])
+    })
+    it("returns primitives and null as-is", () => {
+        expect(clone(5 as any)).toBe(5)
+        expect(clone("hi" as any)).toBe("hi")
+        expect(clone(null as any)).toBe(null)
+    })
+})
+
+describe("checkIfMatching", () => {
+    it("returns true regardless of key insertion order", () => {
+        expect(checkIfMatching({ a: 1, b: 2 }, { b: 2, a: 1 })).toBe(true)
+    })
+    it("returns false when a value differs", () => {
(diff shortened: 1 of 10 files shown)


vassbo:
The functions you added tests for here are already fine-tuned for it's use case so tests like these are not what we need. Tests need to replicate broader functionality, and not one by one function.

Toaster996:
@vassbo I really disagree - how can you e.g. refactor these classes without being sure that they still do what they are supposed to do. I am not saying that they are bad but developing this w/o test really feels like back in the 90s. From what i see, the code or even mutation-coverage of the existing tests is not great.

Toaster996:
Also, of course you needed test that replicate broader functionality. Thats then however not a unit-test, more an integration test. Unit-Test should focus on one thing - one function - and ensure its behaviour.

vassbo:
I don't refactor common functions, or try my best to not do that, many of these are so basic and global that they don't need to change, and should not. Instead if I need some different but similar functionality I create a new function instead.

Toaster996:
I really like this project and hope there will be some more contributions from my side, but I also dont understand this way of developing software. After many years in software engineering, there are mistakes that happen, even if they should not happen and all of them can be traced back to a missing test.

Especially functions used across the project must be bulletproof and extendable. This is all like the core principles of SOLID, right?

vassbo:
Currently it works because there's just one maintainer. And adding tests for all common functions means double work.

If there was a way to add tests directly inside a function instead that could work.

I'm glad you like it, and I appreciate the contribution.

PR #3497, opened 2026-07-09
Fix undo/redo media leak and layout corruption in SLIDES history

Toaster996:
extract action handlers to module-level functions

- Fix media leak: move media mutation into SLIDES history action so undo properly reverts media state (clipboard.ts:663 -> historyActions.ts:472)
- Fix layout corruption: use ID-based layout filtering instead of position- based on undo, fixing multi-slide paste undo (historyActions.ts:525)
- Extract 6 inline action handlers to module-level functions in historyActions.ts (1213->880 lines)
- Extract pasteDrawerItem/cloneChildren helpers in clipboard.ts (1503->1412 lines)
- Remove ~84 lines of dead/commented code across both files

### DIFF
--- a/src/frontend/components/helpers/clipboard.ts
+++ b/src/frontend/components/helpers/clipboard.ts
@@ -550,60 +550,33 @@ const pasteActions = {
         if (!data || get(activePage) !== "edit") return
 
         if (get(activeEdit).id) {
-            if (get(activeEdit).type === "overlay") {
-                if (!get(overlays)[get(activeEdit).id!]) return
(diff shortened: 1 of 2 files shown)


Toaster996:
@vassbo This part looked like it needed some improvement, I think there was also a bug in it. The TODO comment already suggested it.

vassbo:
Thanks, you are correct that it needed some improvements, will review it next week.

vassbo:
It looks good. The current history system is not very good, so one day I would like to improve it to make it simpler.

PR #3498, opened 2026-07-10
Fix stage output window lag/freeze and reduce capture overhead

vicholz:
# Branch: `cfix-video-lag-freeze`

**Commits:** `c0c22313`, `e7879501`, `c1cd458d` (branched from `dev`)

## Summary

Fixes the reported issue where a **stage output window** (a physical output showing a stage layout with an "output window" / `current_output` item mirroring the primary output) would lag over time and eventually **freeze permanently** — especially during video playback — while the primary output and preview thumbnails kept working. Also addresses the **high idle CPU usage and memory fluctuations** observed while the app sat idle.

| Issue | Root cause | Fix |
|---|---|---|
| Stage window lags, then freezes until app restart | IPC listener + component tree leaked on every background/media change in stage output windows | Made listener cleanup symmetric with registration in `BackgroundMedia.svelte` |
| High idle CPU + sawtooth memory | Capture loop ran `capturePage()` at full framerate (20–30fps) forever, even for static content; unchanged frames re-sent every 250ms | Idle backoff to 3fps after 2s of unchanged content; keepalive raised to 1s; loop fully stops when all channels are off |
| Wasted multi-MB IPC to stage windows | Raw capture frames broadcast to local stage windows that never consume them | Removed the broadcast entirely |
| Web stage mirror streamed raw 8MB frames | Poll-based `REQUEST_STREAM` returned full-resolution raw RGBA buffers | Push-based JPEG streaming, only to subscribed mirror viewers |
| Audio pipeline running for muted mirrors | Stage windows attached a Web Audio analysis graph to always-muted mirror videos | Skip analysis when `mirror` is true |
| Permanent high CPU/memory after a StageShow client connects (until app restart) | Connection started output capture; disconnect never re-evaluated it, so the 20fps capture loop ran forever | Capture requires an active mirror viewer, and stops on unsubscribe/disconnect |

### How stage output windows render

A stage output window is an Electron `BrowserWindow` whose output has `stageOutput` set. Its `current_output` item does **not** consume captured pixels — it mounts a full Svelte `Output` component in mirror mode ([Stagebox.svelte:366](https://github.com/vicholz/FreeShow/blob/cfix-video-lag-freeze/src/frontend/components/stage/Stagebox.svelte)):

```svelte
{#if ($outputs[outputWindowId] || $allOutputs[outputWindowId])?.stageOutput}
    <StageLayout outputId={outputWindowId} ... />
{:else}
    <Output outputId={outputWindowId} mirror style="width: 100%; height: 100%;" />
{/if}
```

The mirror re-renders slides/backgrounds itself (including its own muted `<video>` elements, drift-corrected against `videosTime`), so it needs a live IPC receiver for video `TIME`/`DATA` sync.

### Bug 1: the listener leak (the freeze)

Every background media component registers an IPC listener with a **unique id**, and each `receive()` call adds a fresh `ipcRenderer.on(...)` in the preload that only an explicit `destroy(id)` removes. The registration and removal conditions were asymmetric:

```js
// BackgroundMedia.svelte (before)
function startReceiver() {
    const isStage = !!Object.values($outputs)[0]?.stageOutput
    if ((mirror && !isStage) || receiving) return   // stage windows DO register
    ...
    listenerId = "MEDIA_RECEIVE_" + uid(5)          // unique per mount
    receive(OUTPUT, videoReceiver, listenerId)
}

function removeReceiver() {
    if (mirror || !receiving || !mounted) return    // mirror windows NEVER unregister
    destroy(OUTPUT, listenerId)
}
```

In a stage output window `mirror=true` and `isStage=true`, so the component **registers** but `removeReceiver()` always bails on the `mirror` check — on media change and in `onDestroy`. Every background change permanently leaked:

- one `ipcRenderer` listener (every subsequent OUTPUT-channel message fans out to all of them),
- an entry in the preload's `storedReceivers` map,
- the destroyed Svelte component's closure graph, including its detached DOM subtree with `<video>` elements.

Only stage windows leaked (the primary output has `mirror=false` so cleanup ran; main-window previews skip registration entirely), which matches the symptom pattern exactly: primary fine, previews fine, stage window accumulates garbage all session until its renderer effectively dies. Clearing doesn't help because it only updates stores — only an app restart recreates the window.

**Fix** — removal now mirrors registration, and `isStage` only matches actual output windows:

```js
function startReceiver() {
    const isStage = $currentWindow === "output" && !!Object.values($outputs)[0]?.stageOutput
    if ((mirror && !isStage) || receiving) return
    ...
}

function removeReceiver() {
    // remove whenever a receiver was registered (mirrors in stage output windows register too)
    if (!receiving || !mounted) return
    receiving = false
    destroy(OUTPUT, listenerId)
}
```

### Bug 2: capture never idles (idle CPU / memory sawtooth)

When any capture channel is active (NDI, OutputShow, web StageShow, WebRTC), `CaptureLifecycle.runCaptureLoop` called `webContents.capturePage()` at the full channel framerate **continuously**, even for a static slide — allocating and GC-ing an ~8MB image dozens of times per second forever. Change detection only skipped *transmission*, and even that re-sent full unchanged frames every 250ms.

**Fix** — the transmitter already computes a per-frame FNV signature for change detection; it now records when content last changed, and the capture loop reads that to back off:

```ts
// CaptureTransmitter.ts
const signature = this.getQuickSignature(buffer, size)
const changed = !previous || previous.sizeKey !== sizeKey || previous.signature !== signature
if (changed) this.lastChangeTimes[captureId] = now

// CaptureLifecycle.ts — getAdaptiveFrameRate()
// static content - capture at a low rate until a change is detected
if (!options.blackmagic && CaptureTransmitter.getTimeSinceLastChange(id) > this.IDLE_AFTER_MS) {
    return Math.min(baseCaptureFrameRate, this.IDLE_FPS)   // 3fps after 2s static
}
```

Full rate resumes on the first changed frame (worst-case ~333ms latency). Blackmagic is exempt because its frames bypass change detection. The loop now also stops entirely when every channel is toggled off (previously it kept running at 1fps forever), and the unchanged-frame keepalive went from 250ms to 1s.

### Bug 3/4: web stage streaming (subscription-scoped JPEG push)

Previously, web StageShow `current_output` mirrors polled `REQUEST_STREAM` every 500ms and received the **raw full-resolution RGBA buffer** (~8MB at 1080p) per response — and the electron side pushed the same raw frames to all local stage windows (`sendToStageOutputs`) where nothing consumed them. Capture also started whenever *any* stage client connected (even text-only displays) and — a pre-existing bug — kept running after the last client disconnected.

**New design:**

1. Stage clients showing a mirror item **subscribe** with a heartbeat; the app expires stale subscriptions (10s TTL) and cleans up on unsubscribe/disconnect:

```ts
// PreviewCanvas.svelte (stage client)
send("STREAM_SUBSCRIBE", { outputId })
const subscribeInterval = setInterval(() => send("STREAM_SUBSCRIBE", { outputId }), 3000)
onDestroy(() => { clearInterval(subscribeInterval); send("STREAM_UNSUBSCRIBE") })
```

2. Frames are pushed **only to subscribed sockets**, as downscaled JPEGs (~28KB vs ~8MB raw — about 290× less bandwidth), max 10fps, only when content changed:

```ts
// CaptureTransmitter.ts
private static sendFrameToStageClients(captureId: string, image: NativeImage, size: Size) {
    if (getConnections("STAGE") === 0 || getStageStreamSubscriberIds().length === 0) return
    // throttle to 10fps, downscale to max 1280px wide
    const jpeg = frameImage.toJPEG(this.STAGE_FRAME_JPEG_QUALITY)
    toStageStreamSubscribers({ channel: "STREAM_FRAME", data: { id: captureId, jpeg, size, time: Date.now() } })
}
```

3. **Capture itself is gated on viewers** — `shouldBeCaptured()` requires an active mirror viewer (subscriber, or a legacy client still polling `REQUEST_STREAM`), starts on the first one, stops on the last unsubscribe *and* on disconnect:

```ts
// helpers/output.ts
stage: !get(disabledServers).stage && stageConnectionIds.length > 0
    && stageHasOutput(outputId) && hasStageStreamViewers(stageConnectionIds, outputId),
```

Text-based stage displays never subscribe, so they trigger **zero** capture, encoding, or frame traffic — even while videos play. Legacy cached clients keep working through the old raw path, throttled to their own 2fps poll rate (identical freshness, ~10× less IPC). The client keys pushed frames per output id, fixing a clobbering issue when multiple outputs are mirrored.

### Confirmed in isolation: a StageShow connection causes permanently high CPU/memory

This deserves an explicit callout because it is reproducible on its own and matches user reports exactly: **connect a StageShow client once, and CPU/memory stay high until the app is restarted** — even after the client disconnects, and even if the client never displays an output mirror.

Measured on the packaged app (identical scripted test on both builds — 20s baseline, 40s with a stage client connected, 50s after disconnect; per-second sampling of all FreeShow processes):

| Phase | Before this PR | With this PR |
|---|---|---|
| Baseline | 2.9% CPU, flat RSS | 2.1% CPU, flat RSS |
| Stage client connected | **66.2% CPU**, RSS churning 538–582MB — while delivering **zero** frames (client never requested any) | **17.8% CPU** while actively streaming frames to a subscribed mirror viewer |
| After disconnect | **67.1% CPU — permanent until app restart** | **1.8% CPU** — full recovery, RSS stable |

Root cause chain in the old code:

1. `checkWindowCapture()` runs on client **CONNECTION**, and `shouldBeCaptured()` enabled the stage capture channel whenever *any* stage layout contains a `current_output` item (the default layout does) — so merely connecting started the 20fps `capturePage()` loop, whether or not the client ever viewed a mirror.
2. On **DISCONNECT**, `client()` never re-evaluated captures, so nothing turned the capture off.
3. Even when all capture channels were toggled off, the loop itself kept running (enabled-channel state was only checked at start).

This PR closes each link: capture requires an active mirror **viewer** (subscription heartbeat), not just a connection; disconnect and unsubscribe both re-evaluate capture; and the capture loop exits when every channel is off.

### Bug 5: audio analysis for muted mirrors

`$: if ($currentWindow === "output" && videoExists) analyseVideo()` was also true in stage windows, building a Web Audio graph and a 12.5Hz `AUDIO_MAIN` IPC loop for a **muted** mirror video. Mirrors are always muted, so it's now skipped: `$currentWindow === "output" && !mirror && videoExists`.

## Verification

- Before/after lifecycle measurement on the packaged app (see table above): the pre-fix build stays at ~67% CPU with churning memory after a stage client disconnects; this branch returns to a ~2% baseline with stable memory, and uses ~4x less CPU even while actively streaming.
- 20 unit tests (`CaptureTransmitter.test.ts`) covering change tracking, keepalive skip/resend, idle backoff, loop shutdown, push throttling, subscriber gating, and legacy throttling; 8 more for the viewer registry.
- Live socket probe against the packaged app: a connected non-subscribed client received **0 frames** over 10s; after subscribing, ~1fps (static-content keepalive) at ~28KB/frame; **0 frames** after unsubscribe. Frames only began after subscribing, proving capture starts/stops with viewers.
- Verified with `window.api.getListeners()` that the stage window's OUTPUT listener count stays flat across background changes.

---


vassbo:
Did you experience this issue and did you test that it improved the performance?

vicholz:
> Did you experience this issue and did you test that it improved the performance?

@vassbo Yes, I have experienced this many times. It doesn't always happen, and depends on how long FreeShow has been running and what else it was used for, but it does happen. I have not yet been able to test it to confirm that this has fixed the issue. I will try to do some testing Sunday.

vicholz:
@vassbo I was able to replicate the issue on build 1.6.3 and confirm that these change resolve it. The main trigger for this is a StageShow connection. After a connection is started there was higher CPU/memory usage even after the connection is closed. With the fix there is much lower usage, no more lagging or freezing, and the CPU/memory usage dropped back to baseline.

PR #3501, opened 2026-07-10
Add multi-book and comma verse list scripture reference lookup

vicholz:
# Branch: `cfix-scripture-multi-verse-lookup`

**Commit:** `638e9ae7` (branched from `dev`)

## Summary

New feature: the scripture reference search now supports **multiple verse ranges with dashes and commas** and **multiple books separated by semicolons** in a single lookup:

```
Psalm 91:1-3,10;Psalm 23:1-5,8-10
```

selects Psalm 91 verses 1–3 and 10 **plus** Psalm 23 verses 1–5 and 8–10; playing or creating a show includes all of them in order.

| Change | Detail |
|---|---|
| Comma verse lists | `91:1-3,10` works in single references and in every semicolon segment |
| Multi-book segments | `;Psalm 23:…` switches book; segments without a book keep the previous one (existing same-book behavior unchanged) |
| Parser extracted + tested | New `referenceParser.ts` with injected bible data; 11 unit tests |
| Canonical labels | Search box and slide references normalize to e.g. `Psalms 91:1-3+10 ; 23:1-5+8-10`, repeating the book name when it changes |
| Multi-book selection model | Optional per-chapter `books[]` axis through content loading, slide generation, labels, and show creation |
| Latent bug fixed | Slide splitting mapped verses to chapters by chapter *number* — collides when two selected books share a chapter number |

### How scripture lookup worked before

The drawer search parses references through `json-bible`'s `bookSearch`, which understands `Book Chapter:verses` with `-` ranges and `+` lists (`"1-3+5"` → `[1,2,3,5]`) — but **not commas**. A pre-existing `parseMultiChapterReference` handled semicolons, but pinned every segment to the *first* book. The whole selection model was single-book:

```ts
activeScripture.reference = {
    book: number | string,
    chapters: (number | string)[],      // parallel arrays: one chapter per segment
    verses: (number | string)[][]
}
```

Content loading resolved one `Book` and mapped `chapters` through it; slide generation, reference labels, and show creation all assumed that one book.

### Comma support

Commas are normalized to the `+` separator json-bible already understands — only between digits, so a trailing comma while typing is untouched, and the user's typed commas aren't rewritten in the search box (the normalized value is used for parsing only):

```ts
// referenceParser.ts
export function normalizeVerseSeparators(value: string) {
    return value.replace(/(\d)\s*,\s*(?=\d)/g, "$1+")
}
// "91:1-3,10" → "91:1-3+10" → json-bible → [1, 2, 3, 10]
```

The single-reference path in `Scripture.svelte` parses the normalized value too, so `Psalm 91:1-3,10` works without any semicolons.

### Multi-book parsing

The parser moved out of the component into [referenceParser.ts](https://github.com/vicholz/FreeShow/blob/cfix-scripture-multi-verse-lookup/src/frontend/components/drawer/bible/referenceParser.ts) with the bible data injected (`BibleDataLike` — just `bookSearch` + `getBook`), making the grammar unit-testable. Each semicolon segment can now resolve its own book; segments without one keep the previous book:

```ts
let currentBookNumber = firstBookNumber
for (const rawSegment of segments) {
    const segmentBook = getSegmentBookNumber(bibleData, rawSegment)  // 0 for bare "23:1-5"
    if (segmentBook && segmentBook !== currentBookNumber) {
        currentBookNumber = segmentBook          // "Psalm 23:…" switches book
        currentBookName = getCanonicalBookName(books, segmentBook)
    }
    // bare segments get the current book name prefixed so bookSearch can resolve them
    const normalized = normalizeSegment(bibleData, rawSegment, currentBookName, currentBookNumber)
    ...
}
```

The result carries a `books[]` array parallel to `chapters[]`/`verses[][]`. Labels are built from the *parsed selection* rather than echoing typed text (so "Psalm" labels as the canonical "Psalms"), with the book name repeated whenever it changes — including returning to an earlier book:

```
"Psalm 91:1;John 3:16;Psalm 23:1"  →  "Psalms 91:1 ; John 3:16 ; Psalms 23:1"
```

Whole-chapter segments (`;John 3`) and cross-chapter spans (`Genesis 1:30-2:2`, expanded per chapter using each chapter's real verse count) work per-book.

### Threading `books[]` through the pipeline

The selection model gained an optional per-chapter axis — absent for single-book selections, so existing behavior is untouched:

```ts
// stores.ts
reference?: { book; books?: (number | string)[]; chapters; verses }

// getActiveScripturesContent — each chapter can belong to a different book
const chapterBooks = active?.books?.length === selectedChapters.length
    ? active.books : selectedChapters.map(() => active?.book)
const Books = await Promise.all(chapterBooks.map((b) => BibleData.getBook(b)))
const Chapters = await Promise.all(selectedChapters.map((c, i) => Books[i].getChapter(c)))
```

`BibleContent` carries `bookNames?: string[]`, and `buildFullReferenceRange` groups consecutive chapters by book (preserving the user's given order — important because the range joiner sorts chapters numerically, which would otherwise put Psalm 23 before 91):

```ts
// "91:1-3+10 ; Psalms 23:1-5+8-10"  (first book name added by callers)
const includeBook = index > 0 && !!group.book
return includeBook ? `${group.book} ${range}` : range
```

Slide generation, the output reference (`{scripture_reference}` dynamic values), group names, and show creation all pass `bookNames` through.

### Latent collision fix

Slide splitting mapped each selected verse back to its chapter with `chapters.findIndex(c => c == verseContext.chapter)` — by chapter *number*. With two books sharing a chapter number (e.g. `Psalm 3` and `John 3`), verses would land in the wrong book's slot. Verse contexts now carry their exact chapter index:

```ts
type VerseContext = { chapter: number | string; verse: number | string; chapterIndex?: number }
// prefer the exact index (chapter numbers can repeat when multiple books are selected)
const chapterIndex = verseContext.chapterIndex ?? bible.chapters.findIndex((c) => c == verseContext.chapter)
```

### Drawer behavior

The drawer preview shows the **last** segment's book/chapter — the same convention the existing multi-chapter feature uses (it shows the last chapter). The full multi-book selection lives in the reference and is what gets played/output. Manually clicking a chapter or book collapses back to a normal single-book selection.

## Verification

- 11 unit tests in `referenceParser.test.ts` against a faked json-bible: the exact requested example, book switching and returning, bare-segment book inheritance, whole chapters, cross-chapter spans, single references returning to the normal path, incomplete trailing segments while typing, and unknown books.
- Full unit suite, svelte-check (one finding *fewer* than the dev baseline), prettier, and the production frontend build all pass.
- Not yet exercised live end-to-end (the test profile has no bible loaded) — worth checking in a real instance: the example reference, arrowing through output slides across the book boundary, and that plain references like `John 3:16` behave identically.

---


vassbo:
Multiple chapters should be possible. But I don't see the need to select multiple books, if people need multiple books they should select one book at a time, but drag them into the same show as slides.

PR #3520, opened 2026-07-17
Add per-item toggle for zero-padded dynamic timer values

### LINKED ISSUE #3515: Configurable timer dynamic value formatting (zero-padding)
**Describe the feature**
Currently, timer dynamic values are hardcoded with zero-padded minutes:

- {timer_countdown} → 05:07
- {timer_m_countdown} → 05
- {timer_s_countdown} → 07

There's no way to get non-padded output (e.g., 5:07, 5, 7), which matters when you want a cleaner look for smaller numbers. 

In my use case, we transition from a pre-rendered timing video to a dynamic timer configured in FreeShow. While I concede that this is uh... perhaps convoluted, this is the implementation I'm working with. While I can in effect match, it's not perfect because of this zero-padding behavior.

**Workaround**
I've (with local LLM assistance) patched this locally by adding a timer_gp_ variant that strips zero-padding from minutes:

```// showActions.ts — new gp (gross minutes) prefix
let gp = dynamicId.startsWith("timer_gp_")

if (gp) {
    return currentTime >= 60
        ? Math.floor(currentTime / 60).toString()  // no padStart
        : "0"
}
```

Now {timer_gp_countdown} → 5, and you can compose custom formats like {timer_gp_countdown}:{timer_s_countdown} → 5:07.

**Ask**
A few options, in order of preference:

1. Format dropdown on timers (like Clock items already have) — let users pick padded vs. non-padded output.
2. Additional dynamic value variants — e.g., timer_gp_* (gross/unpadded minutes) and timer_gs_* (unpadded seconds), so users can mix and match.
3. Per-timer setting — toggle zero-padding on the timer configuration itself.

devmaiter:
Closes #3515

Adds a per-slide-item checkbox ("Zero-padded timer values") to toggle zero-padding on dynamic timer values, as discussed in the issue.

- The checkbox appears in the text section of the edit panel, only when the item's text contains a `{timer_` dynamic value.
- Default is on, so existing shows are unaffected. Turning it off renders `5:30` instead of `05:30` (also handles single min/sec values, overflow, and timers with hours/days).
- Stored as `dynamicPadding` on the item, following the same pattern as `textFit`.

Tested with `svelte-check` (no new errors in the touched files).

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1056,6 +1056,7 @@
         "transpose_down": "Transpose down",
         "auto_size": "Auto size",
         "no_wrap": "Text on one line",
+        "padded_numbers": "Zero-padded timer values",
         "line_spacing": "Line spacing",
         "line_radius": "Line radius",
         "line_height": "Line height",
--- a/src/frontend/components/edit/tools/BoxStyle.svelte
+++ b/src/frontend/components/edit/tools/BoxStyle.svelte
@@ -266,6 +266,8 @@
         setBoxInputValue(box, "default", "font-family", "styleValue", getStyles(style)["font"] || "")
         // setBoxInputValue(box2, "default", "textFit", "hidden", !item?.auto)
         setBoxInputValue(box, "text", "nowrap", "value", !!styles["white-space"]?.includes("nowrap"))
+        setBoxInputValue(box, "text", "dynamicPadding", "value", item?.dynamicPadding !== false)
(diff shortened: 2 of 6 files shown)


vassbo:
No sorry, on the "Timer" item, not Dynamic values.

vassbo:
Something like this: https://github.com/ChurchApps/FreeShow/commit/c78ad01553f2bfe8237ef43bc4ec9b0b829efa2c

PR #3548, opened 2026-07-27
Add document-style chord editor

gladsonsam:
Just an idea I had when I saw that we have three editing modes. Would be nice to implement something like this where we can move chords around and change keys easily, as well as to reimport the song from chord sheets instead of having to delete and reimport the song.


vassbo:
Thanks, but this is not better than the current way. And we have the Text edit for changing the chords across the entire song.

PR #3551, opened 2026-07-28
Fix: httpsRequest never aborts on timeout, can hang cloud sync forever

lucaxs09:
httpsRequest() sets a timeout option on the request, but never listens for the timeout event. In Node, that option alone only emits the event, it doesn't abort the connection. If nothing calls .destroy(), a hung/slow response leaves the callback (and anything waiting on it, like cloud sync) unresolved forever.

Fixed by destroying the request on timeout. Also added a small idempotency guard, since a timeout that fires after the response has already started can trigger both the request's and the response's error handlers, which would otherwise call the callback twice.

Confirmed the hang is real: reproduced it against a server that never responds (callback never fires without the fix, resolves at ~10s with it), and saw the app itself hang on "Downloading data..." during testing related to #3494.

Added two tests covering both cases.

### DIFF
--- a/src/electron/utils/requests.test.ts
+++ b/src/electron/utils/requests.test.ts
@@ -1,0 +1,81 @@
+import { EventEmitter } from "events"
+import { describe, expect, it, vi } from "vitest"
+
+vi.mock("../IPC/responsesMain", () => ({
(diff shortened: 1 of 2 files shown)


vassbo:
Great! Thanks a lot. :)

Issue #3434, opened 2026-06-21
Cloud Sync with ChurchApps not working properly


quanyan:
**Describe the bug**

After created a show the show able to sync to cloud. On the other PC could't sync the show to the show and the project drawer show Unnamed.

**Reproducing**

PC1 create a new show and drag to the drawer then close the freeshow 
PC2 open freeshow the drawer show Unnamed and could not see any new Shows on the library
working fine on 1.5.9

**Screenshots (Optional)**

<img width="268" height="359" alt="Image" src="https://github.com/user-attachments/assets/f70077c9-5b26-4713-a448-4eb04176670d" />

**Version (Optional)**
 - OS: 
 - FreeShow Version: 
 1.6.2

**Additional context**

KimemiaJ:
This happened to me too after updating to 1.6.2

vassbo:
@lucaxs09

vassbo:
I assume you have 1.6.2 on both computers?

KimemiaJ:
Yes 1.6.2 on both the computers.

vassbo:
I think I figured it out, will release an update in a couple of days.

lucaxs09:
Was this related to the per-item sync merge I added in 1.6.2, or something separate? I'll try to reproduce it and look into it. @vassbo did you already find the cause?

vassbo:
I changed it from merging all the .show files into a single .json to just having them as individual files in the zip, because it seems like trying to merge into a .json sometimes didn't work for many .show files, and also sometimes randomly did not include all the .show files. So it I think it was caused by that, I hoped someone would catch any issues in the beta.

I think I got it fixed, but you could give it a test if you would like. :)

lucaxs09:
Thanks for the explanation, that makes sense. Having them as individual files in the zip sounds more robust than merging everything into one json. I'll do some testing and let you know how it goes.

frederickjh:
Not sure if this is related @vassbo but I have been seeing that adding notes to a show  are not syncing even though the show did sync. It is like it is not seeing the notes as a change.

I do think this is related. For Sunday service I added in the preacher's desired scripture passages as shows at home to the project I had created at church earlier in the week, but when I got to church the shows had synced (were in the project) but were not found. Thankfully I also had exported the project to a USB stick and was able to import it and the scripture shows.

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.6.3

[...]

PR #3556, opened 2026-07-30
Fix: full-file stores (EVENTS/THEMES/MEDIA) can lose real edits to a stale re-sync

lucaxs09:
I found that when syncing full-file stores (`EVENTS`, `THEMES`, `MEDIA`), the comparison to decide which copy wins is based on file `mtime`. The problem: a device's local file mtime gets bumped to "now" by the sync's own download write, not by an actual edit. That means a device that merely re-syncs (without editing anything) can end up looking newer than another device's real edit that hasn't reached the cloud yet, silently discarding that real edit.

This can explain the background brightness report in #3434 (levels reverting to default after syncing): brightness lives in `MEDIA`, one of these full-file stores, and has no per-item `modified` field, so it's a different root cause than the earlier "notes" timestamp bug reported in the same thread.

Note: I haven't reproduced this live end-to-end yet, I haven't had the time. I'll do it as soon as I can, but if anyone else can give it a try in the meantime, it'd be appreciated.

Fix: track the real edit time separately (`Changes.fileModified`, additive field in `changes.json`), bumped only where genuine edits are persisted (`save.ts`), and prefer it over file mtime whenever either side has one. Falls back to the old mtime comparison when neither side has ever tracked an edit (backward compatible with older clients).

**Known limitation, disclosed on purpose:** during a rollout with mixed client versions, an old client that never adopted this field can have its own real edit discarded in favor of an older tracked value from an updated client — there's no way to tell "a harmless re-sync" from "a real edit by a client that doesn't track this" without that client tracking it too. This is the accepted cost of fixing the (much more common) reported bug for fully-updated fleets. Pinned down explicitly with a test so it's a conscious tradeoff, not a hidden one.

Added 3 tests covering: the fix itself, backward-compatible mtime fallback, and the known limitation above.

### DIFF
--- a/src/electron/cloud/syncLedger.ts
+++ b/src/electron/cloud/syncLedger.ts
@@ -1,10 +1,8 @@
 // ----- FreeShow -----
-// Pure, testable per-item merge ledger for SYNCED_SETTINGS / collections.
+// Pure per-item merge ledger for SYNCED_SETTINGS / collections.
 //
-// This extracts the created/deleted ledger decision logic from syncManager (see #3335) into a
-// self-contained unit with NO module globals and NO I/O: all state (changes, cloudChanges, the
-// current deviceId, device count) is provided to the constructor, so the merge decisions can be
-// unit-tested in isolation. syncManager delegates to this instead of keeping the logic inline.
+// Extracts merge logic from syncManager to be unit-testable in isolation.
+// All state is provided to constructor, with no module globals or I/O.
 
 export type Changes = {
     version: string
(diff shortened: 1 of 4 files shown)


lucaxs09:
Note: I noticed the `dev` branch doesn't exist anymore, so I targeted this PR at `main` instead. Just checking, was that intentional?

vassbo:
Because I'm building a release at this moment, I guess this can also be included

lucaxs09:
By the way, I'm also working on improving the generic "Sync Failed" error message so it actually tells the user what happened (right now it never carries any detail about the real cause). If it's useful, I might have something ready in time for this release too.

vassbo:
I suppose you can send a PR for that, but I that will have to be for the next release.

lucaxs09:
Okay. Yes, as soon as it's ready I'll send the new PR

Issue #3512, opened 2026-07-13
投影片編輯錯誤


thomassun920320-eng:
**Describe the bug**

當輸入完中文字後按下enter字會重複出現，造成輸入錯誤
**Reproducing**

進入投影片編輯的頁面打字，選擇中文，選完字按enter或單純按enter都會觸發
**Screenshots (Optional)**

<img width="1710" height="1107" alt="Image" src="https://github.com/user-attachments/assets/ae3c4f8a-9b7c-49b5-9f57-ed5a9d447f03" />
<img width="1710" height="1107" alt="Image" src="https://github.com/user-attachments/assets/eee4956d-253b-4f7c-b68f-dd8470712f6d" />
<img width="1710" height="1107" alt="Image" src="https://github.com/user-attachments/assets/6c4207e4-4e17-4a00-a6d8-bda65b06783b" />

**Version (Optional)**
 - OS: 26.5.2
 - FreeShow Version: 1.6.2

**Additional context**

vassbo:
I can't reproduce unfortunately. Any way to get the same behaviour with English characters?

thomassun920320-eng:
經過測試英文無法重現bug，不知道我還能為您提供什麼幫助

vassbo:
> After testing, the bug cannot be reproduced in English. I don't know what else I can do to help you

Okay, I can't help you then, sorry.

PR #3562, opened 2026-07-31
Don't rewrite the editbox while an IME is composing

### LINKED ISSUE #3553: IME (Chinese/Japanese/Korean): committing a candidate with Enter duplicates the text in the slide editor — no composition event handling anywhere
### Describe the bug

When typing with an IME (Chinese / Japanese / Korean — anything that uses a candidate window) in the slide text editor, committing a candidate duplicates the text. Typing `你好` and pressing <kbd>Enter</kbd> to confirm the candidate leaves `你好你好` in the textbox.

This was reported in #3512 and closed as *can't reproduce*, because it cannot be reproduced with English characters — by definition, since English typing never enters an IME composition session. I did some digging in the source and I believe I've found why. Details below, plus a reproduction recipe that doesn't require knowing any Chinese.

### Reproducing

The only prerequisite is an IME. Both macOS and Windows ship one for free and it takes about a minute to enable — **you don't need to read Chinese to reproduce this**, `nihao` → <kbd>Enter</kbd> is enough.

**macOS:** System Settings → Keyboard → Text Input → Input Sources → Edit… → **+** → Chinese, Traditional → *Pinyin — Traditional*. Switch input source with <kbd>Ctrl</kbd>+<kbd>Space</kbd>.

**Windows:** Settings → Time & language → Language & region → Add a language → 中文 (繁體，台灣) → add the *Microsoft Pinyin* keyboard. Switch with <kbd>Win</kbd>+<kbd>Space</kbd>.

Then:

1. Open a show, enter the slide editor, click into a textbox.
2. Switch to the Pinyin input source.
3. Type `nihao` — a candidate window appears above the caret with underlined provisional text in the textbox.
4. Press <kbd>Enter</kbd> (or <kbd>Space</kbd>, or click a candidate) to commit.
5. The committed characters appear **twice**.

A Japanese IME (`konnichiha` → <kbd>Space</kbd> → <kbd>Enter</kbd>) or a Korean one behaves the same way, so this is not Chinese-specific — it affects every CJK user.

### Root cause (my analysis — worth verifying, but the key evidence is easy to check)

The one thing I can state as fact: **there is no IME composition handling anywhere in the codebase.** Searching the whole repo for `isComposing`, `compositionstart`, `compositionupdate`, `compositionend` or `keyCode === 229` returns zero hits outside `node_modules`. Every text-editing path treats composition input as if it were ordinary finished input.

That matters because of how the editbox round-trips its content. In [`src/frontend/components/edit/editbox/EditboxLines.svelte`](https://github.com/ChurchApps/FreeShow/blob/main/src/frontend/components/edit/editbox/EditboxLines.svelte):

- L704 — the contenteditable uses `bind:innerHTML={html}`. Svelte 3 implements that binding by listening to the **`input`** event.
- L106-110 — whenever `html` changes, `updateLines` is scheduled on a 10 ms timer.
- L415+ `getNewLines()` — reads the live DOM back into the `lines` model and writes it to the store.
- L118-122 `refreshStyleAndRestoreCaret()` → L96 `getStyle()` — regenerates `html` from the model, which **reassigns `innerHTML`**, and then calls `setCaret()`, which does `removeAllRanges()` / `addRange()` on the document Selection.

During an IME composition the browser fires `input` events for each provisional (pre-edit) update, with `isComposing === true`. Nothing filters those out, so the sequence above runs *while a composition is still active*:

1. The provisional text is parsed out of the DOM and committed to the model as if it were real content.
2. `getStyle()` replaces `innerHTML`, destroying the text node the IME had attached its composition to, and `setCaret()` moves the selection out from under it.
3. When the user then commits, `compositionend` inserts the final string into the (rebuilt) DOM — on top of the provisional text that was already persisted in step 1. Result: doubled characters.

There is a second, independent problem on the same key. L602-605:

```ts
function textElemKeydown(e: KeyboardEvent) {
    if (e.key === "Enter" || e.key === "Backspace" || e.key === "Delete") {
        recentKeyboardLineMutationAt = Date.now()
    }
```

The <kbd>Enter</kbd> that commits an IME candidate is **not** a line break — it never reaches the document as a newline, and it arrives with `isComposing === true` (`keyCode 229`). Treating it as a line mutation flips `keyboardLineMutation` for 250 ms and sends `getNewLines()` down a different caret-restore branch (L510-512, L560-561) at exactly the moment the composition commits. The same applies to the global `keydown` handler at L124-156, whose <kbd>Enter</kbd> + <kbd>Alt</kbd> split path also runs during composition. <kbd>Backspace</kbd> is affected too — in most IMEs it deletes one character of the pre-edit buffer rather than any committed text.

I want to be straight about confidence: the missing composition handling is verifiable by grep, and the round-trip above is plainly not composition-safe. I have not stepped through it in a debugger to confirm the exact duplication path, so treat the step 1-3 mechanism as the most likely explanation rather than a proven one.

### Suggested fix

The standard approach for a contenteditable that re-renders itself is to freeze the round-trip for the duration of the composition:

1. Add `on:compositionstart` / `on:compositionend` to the contenteditable at L695-708 and keep a `composing` flag.
2. While `composing` is true, skip the reactive `updateLines` scheduling (L106-110), and skip `getStyle()` / `setCaret()` entirely — never touch `innerHTML` or the Selection mid-composition.
3. On `compositionend`, run `updateLines()` **once**, after the browser has committed the final text (a `tick()` or `requestAnimationFrame` is usually needed, as `compositionend` fires before the DOM settles in some browsers).
4. In `textElemKeydown` (L602) and `keydown` (L124), bail out early on `e.isComposing || e.keyCode === 229` so a candidate-committing <kbd>Enter</kbd>/<kbd>Backspace</kbd> is never mistaken for a line edit.

Once composition events are handled, a regression test can be written without an IME by dispatching synthetic `compositionstart` / `compositionupdate` / `compositionend` events.

I'm happy to put a PR together and test it against Traditional Chinese Bopomofo/Pinyin and Japanese input on macOS if that's useful — just let me know if you'd prefer a different shape for the fix.

### Version

- OS: macOS (also reported on macOS 26.5.2 in #3512)
- FreeShow Version: 1.6.4 (analysis against `main` @ 5dc7849); originally reported on 1.6.2

### Additional context

Related: #3512 (original report, closed as can't-reproduce), #3511 (also from a Chinese user, same editor area).

hansli112:
Fixes #3553 (and #3512, which was closed as can't-reproduce for lack of an IME).

## The bug

Typing Chinese, Japanese or Korean in the slide editor duplicates the text — `nihao` + <kbd>Enter</kbd> leaves 你好你好.

Nothing in the codebase looked at composition state: `isComposing`, `compositionstart` and `compositionend` appear nowhere. So while the IME candidate window was open:

1. `bind:innerHTML` fired `input` for every provisional keystroke (Svelte 3 implements that binding on the `input` event).
2. The reactive block read that half-finished text back into `item.lines`.
3. `getStyle()` regenerated the HTML and reassigned `innerHTML`, detaching the text node the IME had attached its composition to, and `setCaret()` moved the selection out from under it.
4. On commit, `compositionend` inserted the final text again — on top of the provisional text already persisted.

## The fix

Freeze the DOM ↔ model round trip for the duration of the composition:

- track composition with `compositionstart`/`compositionend`, releasing on the next task so engines that commit *after* `compositionend` have settled, and cancelling a pending release when an IME commits part of its buffer and immediately composes again (Pinyin partial commit, Japanese segment-by-segment conversion)
- reset the flag on `blur`, so a composition interrupted by a click can't leave the editbox permanently not saving
- skip both scheduling *and* running `updateLines`, and skip `getStyle()`, while composing. Those reactive blocks read `composing`, so they re-run and pick up the committed text on release
- **move the html block above the style-hash block.** Reactive statements run in source order and `getStyle()` assigns `previousHTML`; with the old order, the style block could run first on release and make the html block's condition false, dropping the composed text entirely
- ignore Enter/Backspace as *line* edits while composing — they belong to the candidate window. The guard is deliberately not an early `return`, so <kbd>Ctrl</kbd>+<kbd>V</kbd> still reaches `EditboxPaste`
- ignore composing keys in the global shortcut handler and in `Editbox.svelte`'s own handler, so <kbd>Space</kbd> picking a candidate can't also advance the slide and <kbd>Escape</kbd> dismissing the candidate window can't blur the editbox

## Reproducing it (no Chinese required)

You need an IME; macOS and Windows both ship one and it takes about a minute — `nihao` + <kbd>Enter</kbd> is enough.

**macOS:** System Settings → Keyboard → Text Input → Input Sources → Edit… → **+** → Chinese, Simplified → *Pinyin — Simplified*. Switch with <kbd>Ctrl</kbd>+<kbd>Space</kbd>.
**Windows:** Settings → Time & language → Language & region → Add a language → 中文 (简体，中国) → add the *Microsoft Pinyin* keyboard. Switch with <kbd>Win</kbd>+<kbd>Space</kbd>.

Then: slide editor → click into a textbox → switch input source → type `nihao` → a candidate window appears with underlined provisional text → press <kbd>Enter</kbd>. On `main` the characters appear twice; with this branch, once.

## Testing

`npm run test:unit` passes (64 tests, 4 new). `test:format` clean on the touched files. `svelte-check` reports 224 errors on this branch and 224 on `main` — no new ones. eslint likewise unchanged.

The unit tests only cover `isComposingKey`; the gate and the flag lifecycle have no automated coverage, because `vitest.config.ts` uses `environment: "node"` and there's no jsdom or `@testing-library/svelte` in the tree, so a component test would need a new devDependency. Rather than guess, I verified the reactive semantics by compiling the component with the repo's own Svelte and reading the generated `$$self.$$.update`: `composing` is a tracked dependency of both gated blocks (so they really do re-run on release, and the committed text is not dropped), and after the reorder the html block precedes the style-hash block in the generated update. Happy to add jsdom + a component test if you'd like that dependency.

**To be clear about what is and isn't verified:** the reasoning and the compiled-output check above are solid, but I have not yet run this build and typed into it with a live IME — I'm posting it now so it can be reviewed in parallel. I'll confirm here once I have, and @thomassun920320-eng, who reported #3512, may be able to check it sooner. Japanese would be especially useful, since segment-by-segment conversion exercises the partial-commit path hardest.

## Not included

`Table.svelte`'s cell editor is a second contenteditable, but it uses `bind:textContent` and never regenerates its HTML or touches the selection, so composition survives it — it just persists provisional text that the commit overwrites. Left alone deliberately; say the word if you'd rather it were consistent.


hansli112:
Following up on the note in the description — this is now hand-tested with a live IME, not just reasoned about.

Ran a dev build of this branch on macOS with a Traditional Chinese input method and went through the paths the fix touches:

| | result |
| --- | --- |
| type `你好`, <kbd>Enter</kbd> to commit | committed once — the duplication in #3553 / #3512 is gone |
| type a full sentence without pausing | correct throughout, including where the IME commits part of the buffer and immediately starts composing again |
| <kbd>Esc</kbd> mid-composition | dismisses the candidate window, editbox keeps focus |
| <kbd>Space</kbd> to pick a candidate | picks the candidate, does not advance the slide |
| click away mid-composition, then return | text still there and still saving |

The last three are the cases the first draft of this patch got wrong, so they're the ones worth re-checking if anyone else builds it.

Still unverified: **Japanese**, where segment-by-segment conversion leans hardest on the partial-commit path, and **Korean**. If anyone has either installed, that's the useful next check — @thomassun920320-eng, this is the branch for the bug you reported in #3512, if you'd like to try it.

PR #3567, opened 2026-08-01
Feature: webapp server for coediting

ip4368:
Why: Although there are ChurchApps backup nowadays, being able to directly edit on a webapp is really convenient, that even when people are not on a computer with FreeShow installed, they can still create shows and edit. Also this lower the entry point for serving member of the church, i.e. older paster might find it easier when it is available on the browser, and they don't need to handle all the displays, streaming, where they really want to focus on the content of the show/slides

What:
1. A webapp started by command "npm run build:web && npm run build:headless && npm run dev:server" (I will work on the prod)
2. an API that mimic the whole IPC that local Electron has, allowing us to nearly keep it as the same functionality, and relatively low maintenance cost.
3. a fully editable slides that is synced to the server in realtime, and when creating new shows/templates... they can be saved to the server by just doing control+s/cmd+s
4. multiple sessions can all edit at the same time, and they see the edit in realtime, utlising yjs
5. compatible client updates, allow user to connect to a server with desktop client, where the desktop client will save their local session, then reload the window that is connected to the remote server. Which they will have access to all shows/resources saved on the server, along with all the media/audio (a virtual file system is kinda implemented for that)
6. user can disconnect from the server on the desktop app and it will automatically goes back to a local session
7. when in remote connection mode, the IPC API are kinda split into 2 parts, those IPC that are related to show data/templates/... resouces, they are all fetch from the server, where the local/hardware related IPC are continued to be local, allowing projection/other features such as camera/blackmagic/NDI to work even in remote connection mode
8. The server can be guarded by a shared token

What is not done (probably not MVP):
1. User login, with name showing who is editing which slide just like google slides


vassbo:
Thanks, it's a good idea. But I feel like this feature would not be worth maintaining.
There are existing external tools that does pretty much the same, like _Chrome Remote Desktop_.

ip4368:
I think the feature is relatively low maintenance, cuz generally on the desktop client side, there is minimal changes as connecting to a compatible API and the code specific for remote connection is fairly minimal.

Remote desktop allows multiple people to have read access to the shows at the same time, but only one person can edit (or it is a mess). I can imagine in a more complicated productions, there might be multiple people working on the slides, maybe each working on different part of the project, and even maybe 2-3 people working in the same show. Having realtime collab actually helps in all these cases.

vassbo:
78 new/changed files is a lot. At the moment I don't have time for this, maybe in the future.

ip4368:
I think i will maintain a fork for my team then (not planning to distribute), cuz I think this can be really useful. By doing so, I can have a sense maintainability when merging upstream to the fork, and when you think you have time for this, we can then discuss how to approach it.

vassbo:
All right, that's fine. There was too many different changes to the core system in this PR, like IPC changes that I would do as seperate PRs.

PR #3569, opened 2026-08-02
Feature: Messages - templated audience alerts with {{token}} fill-ins

gabe-almeida:
## What

A new **Messages** feature: reusable templated audience alerts (e.g. `Parent of child {{number}}, please come to the lobby`) that operators define once and trigger live with per-trigger token values — from a new **Functions ▸ Messages** drawer tab, or via the API (`trigger_message` / `clear_message` / `get_messages`) over WebSocket, REST, OSC and Companion, and as action steps.

This fills a gap for teams coming from ProPresenter's Messages panel (nursery calls, parking notices, service alerts) that currently requires hand-editing an overlay mid-service.

## How it works

- **Zero new output plumbing**: triggering a message materializes a transient, marked overlay (`fromMessageId`) so it rides the existing overlay output layer — transitions, multi-output, stage mirroring and `placeUnderSlide` semantics all come for free. The transient overlay is hidden in the overlay drawer and removed again on clear.
- `{{tokens}}` resolve at **trigger time** (per-trigger values > message defaults); single-brace `{dynamic values}` are left for **render time**, so live values (timers etc.) keep ticking inside a shown message.
- Optional auto-clear after `displayDuration` seconds; re-triggering updates the shown message in place (no stacking).
- `messages` is a synced setting: save/load, cloud per-item merge (`SYNCED_SETTINGS_COLLECTIONS`), undo/redo via history stores, profile access gate (`getAccess("messages")`).

## Testing

- New unit tests (`messages.test.ts`, 11 cases) cover token resolution, materialization, re-trigger replacement, output lock, auto-clear + cancellation, clear-and-delete, closest-name trigger, and the getter — full suite passes (76/76).
- `svelte-check` and the electron `tsc` config pass with no new errors versus a clean baseline of `main`.

Happy to adjust naming, placement, or scope to fit the project's direction.


vassbo:
This is already a thing with Text Variables!

PR #3572, opened 2026-08-03
Fix output capture resolution ignoring configured size on HiDPI displays

schplay:
Currently FreeShow renders every output as a real BrowserWindow and grabs frames with webContents.capturePage(), which returns device pixels. On a scaled display (e.g. a 4K primary at 150%), a configured 1920×1080 output was captured (and sent over NDI) at 2880×1620. This ignored the configured output resolution and pushed 2.25× the pixels through every stage (capture → toBitmap → change-detection → IPC → encode), which also throttled output frame rate.

To correct this, we now render capture-bound output windows at bounds ÷ displayScaleFactor so that logical × scaleFactor === the configured resolution, making capturePage() return exactly the configured size. This corrects the resolution at the source rather than downscaling after capture.

The correction is keyed on capture/hidden state, not the invisible flag. It applies to any output that isn't currently shown on a physical display (invisible outputs and enabled-but-not-displayed outputs). When an output is shown on a display, it fills that display as before (avoiding double-scaling with the existing getSecondDisplay handling).

Testing
  - On a 150%-scaled display, an NDI output configured at 1080p is now received at 1920×1080 (was 2880×1620), verified with an external NDI receiver (such as OBS or NDI Studio Monitor).

A positive side-effect is that this decreases the amount of pixels being processed for these outputs meaning that the effective FPS being output improves.


vassbo:
Great, thanks!

Issue #3547, opened 2026-07-26
Auto Scripture from Audio Input


John-Afolayan:
**Describe the feature**
I recently learned about [Pewbeam](https://pewbeam.com/) and thought it would be cool to add a similar feature to Freeshow. A speaker would say a bible passage and Freeshow could automatically display said passage.

**Additional context**
High-level, this would require:
1. An audio input source from the microphone.
2. A local (or cloud) real-time transcriber such as OpenAI's Whisper or Deepgram.
3.  If it doesn't already exist, Freeshow APIs that allow for a bible passage to be inputted and displayed. As well as support for specific bible versions

Take a look at the [Pewbeam repo](https://github.com/appdever01/pewbeam) for more context.

vassbo:
Would be nice. Looks like @AlloDel might have something going.

josephomills:
This would be groundbreaking! 
Another component would be to have voice commands to control scripture. eg next verse, verse X (with a context of an open chapter),, next chapter, XXX translation (switching to XXX translation if available).

vreykin:
Yeah good idea, love the concept. I would hope it won't be locked to just the English language, and that whisper took in context clues to go to the next verse or open to a specific verse. The issue is that context buildup might be a problem especially when a sermon is long. Not a AI expert but these are my concerns. Also I was working on this myself, glad to see someone else also looking for a similar solution.

josephomills:
I've built the AI auto-scripture feature on PRs #3578 & #3579. Need feedback @John-Afolayan and co
@vassbo kindly check it out.
@vreykin also added multi-lingual support and interpretation mode for dual languages.

Added a few improvements: models on Ollama, scripture detection and auto-projection. So everything can be run locally with whisper with or without Ollama models to complement. Or online with Claude, Gemini & OpenAI models.

AlloDel:
@josephomills thanks for pushing this forward on #3578 and #3579. Glad to see Auto Scripture moving. The voice command ideas you raised earlier and the field testing notes are really helpful.

I have also been working on this for a while (as @vassbo mentioned). I have been doing a lot of live testing and experimenting with lighter approaches. My focus has been a bit different:

**Fully local, no API keys for the core path.** FreeShow is open source, and I would love to keep this feature that way too. Mic audio and scripture detection stay on the device. Nothing leaves the machine, and no cloud provider should be required for the main experience.

[...]

PR #3579, opened 2026-08-06
Smart AI STT Auto Scripture & LLM APIs

josephomills:
> Builds on #3578 (its commit is included here) — happy to rebase once that lands, or fold both together if preferred.

## What

**AI auto scripture** (opt-in, off by default): FreeShow listens to the sermon, transcribes locally with whisper.cpp, detects scripture references — both spoken ("John chapter 3 verse 16", instant local matching) and quoted ("for God so loved the world…", via a bring-your-own-key LLM) — and either projects them automatically or offers one-click suggestion cards in the scripture drawer.

https://github.com/ChurchApps/FreeShow — demo video to follow.

## Features

- **Live panel in the Scripture drawer**: status, live transcript ticker (expandable), suggestion cards with confidence badges, restore-previous, manual-override auto-pause
- **Fully local option**: a Local (Ollama) provider — Gemma / Llama / Qwen / Mistral / Phi on the user's own machine, no key, nothing leaves the device
- **Two-tier detection**: instant local matching for explicit references (works with zero network/keys) + optional LLM detection of quoted/paraphrased verses (Anthropic / OpenAI / Gemini via plain REST, user's own key)
- **Quality gates**: confidence-gated auto-projection (quoted verses default to suggestions), quoted matches verified against actual verse text across the user's selected translations, dedupe/cooldowns, music segments (♪) excluded from detection
- **Voice commands** (separate opt-in): "go to the next verse", "give me verse five", "give me NIV", "give me another translation" — imperative-only, and only while a scripture is live
- **Interpretation mode**: per-window language auto-detect for services with live interpreters, constrained to declared languages, with a "detect scriptures from" selector; multi-language transcript tagging
- **Multi-translation search** with select-all, display-translation choice (drawer selection vs matched translation)
- **Session context**: bare "look at verse twelve" resolves against the passage currently live
- Dedicated **AI Scripture settings tab**, guided whisper install per platform, model downloads with progress/resume

## Guarantees (please review these first)

- **Off by default**: no UI, stores, audio capture, downloads or network activity unless explicitly enabled; mic starts only on the operator's click
- **Privacy**: audio never leaves the device (transcription is fully local); only short transcript text windows go to the user's chosen LLM provider, only while listening with a key saved; keyless mode makes zero network calls; the transcript is memory-only
- **Keys**: stored in the ACCESS store (non-portable, never cloud-synced — same as the Canva credentials precedent); raw keys never reach the renderer (booleans only); flagged for a possible `safeStorage` follow-up
- **Zero new npm dependencies** (axios + yauzl already present; whisper.cpp is a runtime download/detection)
- **Supply chain**: Windows binary auto-download is SHA-256-pinned against the official whisper.cpp release; macOS/Linux use system installs (brew/package manager) or a user-picked binary — no unofficial mirrors

## Platform matrix

| Platform | Whisper |
|---|---|
| Windows x64 | official release auto-download (checksum-verified) |
| macOS (Intel + Silicon) | `brew install whisper-cpp` detection or custom path |
| Linux | package manager detection or custom path |

## Testing

- 141 unit tests (vitest, colocated) over detection, transcription windowing, providers, whisper management, voice commands
- Field-tested live at a conference on macOS. Honest note: transcription accuracy with a **laptop mic capturing room audio is limited** — a direct feed from the sound desk (any input selectable in settings) is the intended setup and testing on that is ongoing. Detection precision is deliberately conservative (explicit references only auto-project; everything ambiguous becomes a suggestion).
- Would love testers on Windows/Linux and with mixer feeds.

Happy to split, adjust scope, or rework anything — I kept every touched upstream file minimal and matched existing patterns (LTC worklet, ffmpegManager, ACCESS store, Material inputs, i18n) throughout.

---

## Updates since opening

- **Local Ollama provider** — fully offline quoted-verse detection (Gemma 3/4, Llama, Qwen, Mistral, Phi via ollama.com), same prompt/schema as the cloud providers, no API key; completes the everything-stays-local story
- **Auto-projection fixes from live field testing**: the manual-override pause now only triggers when the operator overrides a scripture the AI itself projected (ordinary song/slide changes no longer silently pause auto mode) and self-resumes after a minute
- **Spoken-form coverage**: "Matthew 12 4", "12, 4", "12. 4", "12-4", word numbers, and cued chapter-only ("turn to Matthew chapter five") all auto-project; a bare "Matthew 5" mid-sentence stays a suggestion
- **Music gating**: whisper wraps singing in ♪ and hallucinates lyrics — music segments are shown faded in the transcript but never feed detection
- **Interpretation hardening**: a declared "languages spoken" set constrains per-window language detection; out-of-set guesses are re-checked against the detection language
- Capture/UX: system-default mic auto-selection, speech-tuned capture (AGC + noise suppression), sample-rate-aware resampling, resumable checksum-verified downloads with live progress

Test suite is now at 141.


josephomills:
Pushed a round of updates from live field testing at a conference, plus a fully-local AI option:

- **Local (Ollama) provider** — quoted-verse detection with Gemma 3/4, Llama, Qwen, Mistral or Phi running on the user's own machine: no API key, free, offline, and nothing ever leaves the device. Cloud providers remain optional alternatives.
- **Auto mode fixes**: the manual-override pause was too eager (any song/slide change paused auto-projection until manually resumed) — it now only triggers when the operator overrides a scripture the AI itself projected, and self-resumes after a minute.
- **Real spoken forms auto-project**: preachers rarely say the words "chapter" and "verse" — "Matthew 12 4" (and every punctuation whisper produces for it: "12, 4", "12. 4", "12-4", word numbers) plus cued chapter-only ("turn to Matthew chapter five") now project automatically; a lone "Matthew 5" mid-sentence still asks first.
- **Music never triggers detection**: whisper hallucinates lyrics for songs it hears (♪) — those segments show faded in the transcript but are excluded from detection and the LLM.
- **Interpretation mode**: declared languages constrain whisper's per-window guess; out-of-set windows are double-checked against the detection language.
- Plus capture-quality fixes (auto mic selection, speech-tuned gain/noise settings, sample-rate-aware resampling) and resumable, checksum-verified model downloads.

141 unit tests passing. The PR description's update section has the full list.

AlloDel:
Left a longer note on #3547 about complementary fully local streaming work I have been testing (no API keys / no LLM for the core path). Happy to collaborate or wait for maintainer preference so we do not step on this review.

josephomills:
@AlloDel your driver is merged and wired up — engine selector in settings, resumable model download, packaging config. Thank you, the VAD choice and the Electron notes were spot on.

One change to your driver, after live testing. The decoder deafness you documented goes deeper than stream reuse: any chunked feeding (fresh, reused, or persistent stream) intermittently decodes short utterances like "next verse" to nothing. The same samples in one batch acceptWaveform on a fresh stream decode correctly every time. So the driver now buffers each utterance (1s pre-roll) and batch-decodes when the VAD closes, waiting 0.5s of real tail audio first so a soft last word isn't cut mid-word. VAD retuned: threshold 0.3, min-silence 0.8s, min-speech 0.15s. If your live-testing notes contradict any of this, I'd like to know.

What I'd like you to review: downloading the sherpa runtime on demand — finishing your zero-setup goal. Right now the npm natives ship in the installer (~16 MB/platform) for everyone. The plan, mirroring our ffmpeg/whisper downloaders:

Exact-pin sherpa-onnx-node@1.13.4; exclude the platform packages from electron-builder, keep the JS wrapper bundled
Self-hosted per-platform ZIPs with pinned SHA-256s (plus proper license files — the npm tarballs omit them)
Extract to userData/bin/sherpa/ with sherpa-onnx-node/ copied beside it, so addon.js's own ../sherpa-onnx-<platform> probe resolves — no patching
macOS: clear quarantine + ad-hoc codesign after extraction (darwin-x64 ships its .node unsigned)
One "Enable" gate chains runtime (~7 MB) + model (662 MB) downloads
Questions:

Does the sibling-layout require() hold across sherpa-onnx-node versions you've used?
Anything from your live testing this misses?
Any concern with self-hosting the natives vs keeping them npm-managed?
Whisper stays the default; settings, multilingual and detection are unchanged.

vassbo:
I will review this soon.

But just a couple things for now I would like changed:
- Can all the new locale strings be placed under their own parent key(s) at the bottom in `en.json`
- Can everything AI related have their own parent folders, so `electron/ai/...` & `frontend/ai/...`

josephomills:
Alright @vassbo 
Will do

josephomills:
Hi @vassbo 
I pushed the changes you requested yesterday for your review, whenever you can. Thanks

vassbo:
I made some changes.
- Can the providers/models setup be general and not scripture specific, because the same models might be useful for other things. - That includes all the IPC messages should not contain `SCRIPTURE`, and most of the locale keys also does not need `scripture` in the key name.
- Can we organize the code into more folders/files.

josephomills:
Okay. Your changes were helpful. 
I've also made the changes as requested.

vassbo:
I essentially have to rework the entire system because your implementation is too specific, and not easily expandable. I already made the biggest changes (very roughly), but there's still lots of things to do, now it probably doesn't work properly anymore, sorry about that but it had to be done.

The system has to be for other things as well, not just scriptures. If you check it out you'll see that I have also added a "bubble" where I would like AI interactions to live.

If you see where I'm going, would you mind porting the old scripture system logic over to the new system? If not I might come back to it next week.

A little more context:
- We should be able to manage engines/models/providers seperately
- Have a toggle for auto scripture, with more specific options like we have
- Later have more toggles for other auto features, like a help to create templates, change settings, etc., and eventually auto lyrics
- One or more models should run in the backgorund depending on what features the user has turned on
- If scripture mode is on and the model is confident it hears a scripture, the "bubble" should show a suggestion to present it in the bottom right corner (or present automatically if that's on)
- We should also be also to only use speech to text to auto find scriptures, even though it might be less accurate, it's faster

josephomills:
True, my implementation is specific cos I am trying to solve a specific issue. I wasn't really setting out with the whole AI infrastructure in mind. So your rework is very helpful to know the general foundation and pathway for AI features. Thank you. 

I've ported the scripture system onto your rework. Engines/models/providers are now managed separately (the model manager popup also got the LLM provider/key section you'd stubbed), the auto scripture toggle runs the models in the background, confident detections show as suggestion cards on the bubble bottom-right (or project automatically in auto mode), and speech-to-text-only detection works with no LLM (tier-1 regex + the on-device quote matcher, all suites green — 315 tests).

Along the way I fixed what the rework left broken: engine status now comes from the runtime resolvers (downloadedModels was never populated / nemotron checked a nonexistent binary), the nemotron downloader wrote one level above where the loader reads, cancel/delete/verify were TODOs, and provider errors lost the stable codes the detection coordinator pauses/backs off on — those are restored on the APIModel base. I also re-exported the pure helpers the colocated tests import (they didn't compile without them) — shout if you'd rather restructure the tests.

A few questions: 
(1) What do AI features for the other features/tabs (eg Shows, Media, Functions etc) look like? Asking to better understand the trajectory so I align without the need to rework my implementations
(2) What's the plan for the expanded bubble long-term aside scripture? in terms of UI/UX also? is it just going to have the transcript only?
(3) Depending on the above, can the live transcript preview remain in the expandable panel I have (outside of the bubble)? Cos I'm wondering how it can be absorbed into the bubble and I also think it's important to see the live transcript. Do you prefer it to also float rather, like the bubble but maybe centered?
(4) For the expanded bubble — I'm thinking the user must be able to work on the app while listening (e.g. manually open scriptures the AI missed), so suggestions live on the closed bubble and the modal is an optional detail view. Match your intent?
(5) For scripture auto show mode, can we have a confidence slider for the user to decide what level of confidence before auto-showing? is that an example of the specific options you mentioned?

vassbo:
Thanks.

1. Not sure yet, we'll have to figure it out as we go. Let's do only scripture for now.
2. It's supposed to be used to inform the user about anything the AI say/do, and we can have some settings inside the popup.
3. My thinking is that the latest transcript/suggestion will show up in the bottom right with interactive buttons, without opening the bubble, but if you open it you can get more info. So I'm hoping we can get the same functionality with just the "bubble".
4. Yes, exactly.
5. I guess that makes sense.

josephomills:
@vassbo  Summary of what's implemented on this branch so far:

**Speech to text (offline, on-device)**
- Two engines behind one generic STT layer: **whisper.cpp** (CLI or resident server, 5s windows / adaptive step, seam-stitched so words neither repeat nor go missing, anti-hallucination guards for silence & music) and **NVIDIA Nemotron** via sherpa-onnx (Silero-VAD utterances, streamed partial decodes while a sentence is still being spoken).
- Decoder vocabulary biasing for whisper: biblical names & KJV archaisms, re-biased toward the book currently being preached from.
- Mic + engine can be switched mid-session; default input follows the system default device.

**Scripture detection (works with no LLM at all)**
- Tier 1: spoken references ("John chapter 3 verse 16", "John, 1, 1", "Romans 8 28 through 30", glued "John 316") with spoken-number normalization and confidence rules.
- ASR-mishearing tolerance throughout: the word "verse" heard as "best"/"this", verse numbers heard as homophones ("Matthew 4 *for*"), and mangled book names ("Corinians" → 1 Corinthians, "Philippines" → Philippians) — all constrained to shapes narration can't produce.
- Optional tier 2 LLM detection (Anthropic/OpenAI/Gemini/Ollama) with typed error codes, backoff, and provider "None" as the explicit default — the whole feature runs keyless.

**Quote matching (local, no LLM)**
- Verbatim recitations matched against every installed translation: compact columnar indexes (~3–6 MB per full bible, all translations fit a 256 MB budget), inverted prefix voting + order-aware weighted alignment, anchor hysteresis for parallel passages.
- Follows a reading **verse by verse** (continuations advance the projection as the reader moves), self-corrects a similar-passage mispick once later words settle it, and recovers misheard rare words phonetically ("analekite" still finds the Amalekites).
- Voice commands: "next/previous verse/chapter", "give me verse 5", "show chapter 4", translation switching ("give me NIV").

**Performance & responsiveness**
- Nothing blocks anything: Nemotron decodes in its own utilityProcess, quote matching runs in a Web Worker (bundled inline so packaged `file://` windows load it), whisper stretches its step under load instead of dropping words. Overload degrades to latency — never freezes, never lost text.

**UI**
- One floating AI bubble as the single surface: live transcript ticker, suggestion cards with confidence, pause/resume, settings popup kept feature-agnostic.
- Auto-projection with percentage confidence threshold (low/medium/high bands), per-reference & display cooldowns, output-lock awareness.

In progress: matching from short distinctive fragments ("and there was light") rather than near-full recitations — candidate lookup and scoring changes for search-field-like phrase matching.

josephomills:
Follow-up: short-fragment quote matching has landed, plus responsiveness fixes.

**Quote matching from fragments (no LLM)**
The matcher previously needed a near-full recitation; now a short quoted phrase finds its verse the way typing it into bible search would:
- *"and there was light"* → Genesis 1:3, *"god called the light day"* → Genesis 1:5, *"for god so loved"* → John 3:16, *"the valley of the shadow of death"* → Psalm 23:4, *"I can do all things through Christ"* → Philippians 4:13.
- Three layers: a **bigram candidate route** (adjacent word pairs pinpoint verses whose individual words are too common to vote — sparse, ~2 MB, drawer translation), **spill-clean scoring**, and **phrase-run evidence** (a contiguous ordered run is evidence in itself, with guards measured against real sermon speech: liturgical formulas like "in the name of Jesus" / "glory to God in this place" and testimony chatter all stay silent).
- Floors were tuned empirically against a full 31k-verse KJV plus a false-positive sweep of greeting/liturgy/testimony/announcement speech.

**End-to-end validation**
A TTS-preached sermon run through real whisper.cpp (base.en) at realtime pacing, into real detection + matching: explicit "John chapter 3, verse 16" detected high, John 3:16 matched from the first *half* of the verse (through ASR mangling), 3:17 followed as a continuation, Philippians 4:13 fired from a fragment split across two decode windows. Matcher compute 3–6 ms per segment. The harness also surfaced two live bugs, both fixed: whisper timestamps can hallucinate past their window and poison the overlap dedupe (now clamped), and tier-1 missed "chapter 3, verse 16" when whisper writes the dictation pause as a comma.

**Responsiveness & UI**
- **High-confidence detections project immediately** — only medium waits out the current slide's minimum display time (latest wins). The per-reference cooldown still prevents re-projecting the same passage.
- The closed bubble keeps only the newest three suggestion cards; the expanded popup now carries the full list as compact two-line cards; the live transcript ticker sits beside the bubble on its own row.

AlloDel:
@josephomills thank you for porting #3 and #4, and for pinning that phrasing as a named regression test. I checked the branch and all eight of my cases now behave. Keeping the ordinal rewrite to Psalms only was the right call, my generic version would have read "the 1st John" as John 1, which I had missed.

On the quote matcher, that one is on me. I went quiet because I had other deadlines to meet, so building it yourself was the correct decision. Yours is ahead of mine now anyway, the bigram route and the per translation worker indexes are things mine did not do.

Sorry this is late. Answering your questions properly now, in case any of it is still useful.

**On the batch decode change.** I tried to reproduce it and could not. Two tests:

- Six short commands ("next verse", "next chapter", "previous verse", "verse fourteen", "John three sixteen", "switch to King James"), five repeats each, so 30 runs. Each run decoded the same samples two ways on a fresh stream, once fed in 2048 sample chunks and once in a single batch call. The output was identical in all 30 runs, including the empties. Five runs were empty in both modes and they were all the same clip.
- Nineteen real VAD segments from a session recording. Identical in both modes again.

**What I did find.** The decode is sensitive to how much silence sits in front of the audio. The same "next verse" clip decodes correctly with no leading silence and with 1 second of leading silence, but returns empty with 0.5 seconds. That points at frame alignment or encoder warm up rather than how the samples are fed. One clip ("verse fourteen", 0.8s) was empty in every configuration, which looks like a model limit on that audio rather than a plumbing problem.

So I would describe the cause differently, but your change still helps, because a fixed pre-roll and a consistent tail remove that variance instead of leaving it to chance. Worth knowing my clips are synthetic macOS `say` audio, clean and exactly 16 kHz. Yours are real mic audio in a room. Both results can be true at once.

**On the sherpa runtime download.**

1. The sibling layout holds. `../sherpa-onnx-<platform>-<arch>/sherpa-onnx.node` is the primary path in two places, not one. `addon-static-import.js` tries it first as a static require, and `addon.js` lists it in `possible_paths`. The static require sits inside a try/catch, so excluding the platform package from the build is safe. That is on 1.13.4. I would still pin exactly like you planned, because this is an internal layout and not a documented API, so it could change without a major version.

2. One thing to be careful with: the `.node` binary has `@loader_path` as its rpath, so the dylibs must sit in the same folder as `sherpa-onnx.node`. Copying the whole platform folder covers it, which is what you described. If that folder is ever flattened or split, it fails to load at runtime rather than at install time, which is a hard failure to diagnose.

3. No concern with self hosting, given the pinned hashes and the license files you mentioned. The macOS quarantine and ad hoc signing step is needed. The plan looked right to me.

Small thing, and no problem if it is awkward: when the ported work lands, would you mind adding a `Co-authored-by: AlloDel <prosperdela17@gmail.com>` line on those commits? Purely so the history carries it. Happy either way.

josephomills:
Really glad to have you back on this @AlloDel — and no apology needed; the work speaks for itself.

Your batch-decode measurement is genuinely useful. The leading-silence sensitivity (0s and 1s fine, 0.5s empty) is a sharper explanation than the one I'd written down, and the driver's header now carries your finding — with you as co-author on that commit, since it's your measurement (`ca05565a`). Your "both can be true" framing is right: the fixed pre-roll and trailing pad stay regardless, since they remove the variance whichever way it arises — and it matches something we hit independently, where a 1s pre-roll was load-bearing for opening syllables on real mic audio.

Noted on the `@loader_path` rpath caveat — the platform folder gets copied intact, never flattened, and we'll pin the exact sherpa version as planned since the sibling layout is internal. The quarantine/ad-hoc signing step is in the plan too.

Co-author trailers are on the ported commits as requested — the history now carries you on all of them. Heads-up that adding the trailers rewrote those hashes: the port of #3/#4 is now `0bbbb716`, the extended alias table `20a5fc1e`, and your pinned John 3:16 case `97424a4d` (the closing comments on the fork PRs reference the pre-rewrite hashes).

And thank you for re-checking all eight of your cases against the branch — that kind of verification is worth more than the code.

josephomills:
**Since the last update — bubble & scripture polish**

- The floating bubble now sits properly in the app's layering scheme, and the live transcript ticker wears a thin AI ring so it stays readable over any panel.
- The transcript is selectable, and can be copied from its context menu.
- Toggle semantics are locked down: turning AI off always stops a running listening
  session (the main switch used to leave it running invisibly), and it also turns
  AI Scripture off, since the feature needs the AI layer — while AI Scripture can
  still be turned off on its own. Loaded settings are sanitized the same way, so an
  older save can't wake up in the invalid "scripture on / AI off" state after a
  restart. The scripture options panel also follows changes made from anywhere now,
  instead of keeping a snapshot from when it mounted.

**AI in Quick Search**

The palette (Ctrl+G) has a new **AI** category — its own group header, `ai:` / `ai `
prefix filtering, and clicking the group header filters too:

- **AI** — opens the AI settings page
- **Enable/Disable AI** — the main switch, label follows the current state
- **Manage models** — opens the model manager on the AI settings tab
- **Enable/Disable AI Scripture** — the feature switch
- **AI Scripture settings** — opens the scripture drawer with its options panel
  already revealed (through a new generic `drawer_options` trigger any drawer tab
  can reuse)
- **Start/Stop listening** — only listed while the feature is enabled, following the
  live session state

The entries keep a fixed order, and reopening the palette re-runs the search so the
state-dependent labels never go stale. Terms like `whisper`, `transcription`,
`voice`, `microphone`, `models`, `LLM` or `provider` surface the right entries.

All 373 AI tests stay green, and the svelte-check/lint baselines are unchanged.

@vassbo I'll pause here and wait for you to have a look at it so we know the way forward.
@AlloDel you can also have a look when you have time so you can also give me feedback.

Thank you!

vassbo:
@josephomills The plan is to review this soon, and get it merged as a hidden/experimental feature before the next beta release in a few weeks.

josephomills:
Great! Looking forward to a lot of feedback from different speakers/preachers, different languages & different computers.

I'm also looking up ways to simplify this feature:
Eg a default multilingual speech model and improving STT quality. Already seeing massive improvements with the [rework](https://github.com/josephomills/FreeShow/pull/5).

josephomills:
🎉 Upstream win: all four sherpa-onnx PRs I sent for this feature are merged and released in **sherpa-onnx-node@1.13.7**:

- **Language pinning** — k2-fsa/sherpa-onnx#3891: pin the congregation's language per stream, switchable live
- **Utterance closing** — k2-fsa/sherpa-onnx#3892: the decoder now reports trailing silence, so we know exactly when a sentence is finished
- **Confidence** — k2-fsa/sherpa-onnx#3893: per-token scores to back the suggestion confidence
- **Parallel streams** — k2-fsa/sherpa-onnx#3894: batched decoding, opens the door to interpreted/multi-language services later

One more still in review: k2-fsa/sherpa-onnx#3895 adds beam search + hotwords — correctly spells rare biblical names like Melchizedek and Zerubbabel in testing.

Will definitely implement these in the [STT rework](https://github.com/josephomills/FreeShow/pull/5) (1.13.6 → 1.13.7 is purely additive) and share the measured improvements after more testing. @vassbo @AlloDel

josephomills:
Hey @vassbo — the rework reads much cleaner, especially detection moving to the renderer and the generic suggestion model. A few things went out along the way and I'd rather ask than undo something deliberate.

**Granularity** — the one I'd most like to pin down. In August you asked for the AI code split into more files, so I broke scripture into ~10 modules; the recent cleanups go the other way (ten quote-match modules into one 1,660-line engine, whisper manager/transcriber/wrapper into one file). No strong preference either way — I'd just like a rule I can apply so I stop churning your files. Is it "one file per cohesive subsystem, split only by ownership"? And does that extend to the remaining scripture modules, or leave those as they are?

**Vocabulary biasing** — `bibleVocabulary.ts` / `prompt.ts` are gone. Those were biasing whisper on book names before mishearings ("Philippines" for Philippians) ever reached the parser. Wrong layer, or not worth its keep? Happy to re-wire it renderer-side if it's the former.

**LLM error codes** — `codedError` and the provider quirks came out, but `DetectionCoordinator` still branches on `invalid_key` / `model_not_found` / `rate_limited`. `completeLLM` now only emits `invalid_key` or `server_error`, so tier 2 never disables itself on a bad key and never backs off when a provider rate-limits us — it keeps calling every window. Slimmer codes back, or should I rewrite the coordinator around messages?

**redactSecrets** — went out with the same commit. It stripped `sk-…` / `AIza…` from provider errors before they hit the renderer and the logs (OpenAI 401s echo a partial key). Deliberate, or collateral? A few lines to restore.

**Nemotron partials** — the partial re-decode is gone, so text only leaves the driver when the utterance closes. Those partials emitted real segments, so detection saw words ~1.2s in; now mid-flow speech waits for a pause, up to the 12s VAD cap. Short utterances are fine — it's continuous preaching where it shows. Complexity, or did it misbehave? Can measure before/after if that helps.

**Voice commands** — happy to go LLM-first, but two things I can't see how they land. Today commands need no key and no setup; LLM-only means an API key, or installing Ollama and pulling a model. And tier 2 is gated on `LLM_MIN_NEW_WORDS = 15`, so a two-word "next verse" wouldn't trigger a call until another 13 words arrive — while giving commands their own call means a round-trip per utterance. With partials gone too, that's utterance-close plus gate plus round-trip before a slide moves. Is the thinking that commands ride the existing detection call, or something separate? Not arguing to bring the old matcher back, just want to understand the shape you have in mind.

Happy to pick up whichever of these you want and leave the rest.

vassbo:
@josephomills I'm still not done. But taking a break.

- **Granularity:** 10 files for "one thing" is too much. I merged it back to more easily trim it down, but I'm not done yet, 1500 lines in one file is still a bit much.
- **Vocabulary biasing:** Not sure about this one as it's a lot of strings only for English, so I removed it. But I guess we can have the most common mishearings, this should apply to both Whisper and Nemotron.
- **LLM error codes:** I want all error messages to be direct strings (English only), and not translatable codes, as it's much easier to maintain that way.
- **redactSecrets:** I think it's not needed as the key is already partially redacted.
- **Nemotron partials:** This can be re-added if it's kept simple, but it should work with the new system.
- **Voice commands:** It's too much to maintain, and it would only work for English. LLMs are great at understanding any language.

josephomills:
@vassbo Thanks, that's helpful. No rush.

- **Error codes:** agreed, I'll rewrite the coordinator around messages.
- **redactSecrets:** fair enough, dropping it.
- **Vocabulary + partials:** I'll take both on.
- **Voice commands:** it wasn't English-only — a per-language table (en/es/de/fr/no/pt), ~17 short word lists each, matched against the spoken language plus English. Data rather than code, so it could live in the locale files and go through the translators instead of you.
- Whisper could translate to English, so one English grammar would cover it. Nemotron can't — transducers are monotonic — so once 3.5 lands multilingual it'll emit e.g. Spanish text that still needs matching. That's the case a phrase table covers and translation can't.
- **STT rework:** the sherpa 1.13.7 work I mentioned (multilingual 3.5, language pinning, utterance close, per-token confidence) is ready. Happy to rebase it onto your cleanups — good time to push it here, or would you rather finish the restructure first?

vassbo:
**Voice commands:** Right, I remember now. But still only accounts for some languages. I like the idea, but would still not be good to maintain. Let's get the LLM to work well with voice commands at least. Maybe another time...
**STT rework:** Great! I'll do some more restructuring first.

josephomills:
Voice commands: Alright, LLM will work to an extent. My only concern is the latency and dependence on network connection it'll introduce. Will think of how to mitigate that.

STT rework: Sure. Just let me know when to push

josephomills:
Also wanted to ask if you'll like to maintain having a choice of STT model for scripture specifically (ie. Nemotron & whisper) as it is now, or you'd want to have one moving forward.

I started with the 2 because whisper was multilingual but not a streaming model as Nemotron. But I found **Nemotron 3.5 ASR multilingual** and that's what my STT rework uses. And the I think the 'interpretation mode' feature I had is a bit pre-mature, needs more work.

I can see whisper being useful in non-scripture aspects like on the Audio tab for audio files (where Nemotron won't be ideal). But that'll be an architectural call.

vassbo:
At this point I have basically rewritten everything, like every single line of code. But I hope you see how much easier it is to work with, because AI is very good at overcomplicating things!

Fixed partials, the detection seems pretty good now.
Vocabulary biasing might not be needed with the way the current system works now?

Voice commands: After playing around with it more I see what you mean about the LLM delay. But let's keep it out of this PR at least. Also I hope it might not be needed, as if someone start to read any verse it should detect that pretty quickly anyway.

I changed the LLM logic, but it's not finalized. I'm thinking at the moment it's not needed for local XML bibles, because the direct detection takes care of that (much faster). It might be useful for API Bibles, and of course other things like generating slides, but that will come at another time.

If you found a multilingual Nemotron model we should use that as having interim parts is better. Just leave Whisper for now, but I might remove it if the new Nemotron works good.

It's ready for the STT rework now. :)

josephomills:
Thank you! Will have a look and rebase the STT rework.

On the over-complication, to an extent, yes. But a lot of the removed portions are actually deliberate features and decisions from live tests. 

Eg.
- Multi-bible search - now restricted to the selected translation
- Follow in drawer - moving to AI projected verse
- cool-down for same reference quoted
- verse bounds to protect against projecting non-existing references
- index memory budget & and indexing on worker
- glued reference splitting eg John 116 (when the speaker says  John one sixteen (ie John 1:16))
- detection web worker 
- voice command grammar
etc etc

Each contributing to faster or better detection (in my opinion, lol). the thread a feature runs on is architectural so that's okay (whether main thread vs worker). Then also I'm heavily test-driven that's why there's a lot of tests which I see you've removed almost all. Lol

So I'll test it well and also with a benchmark I have to see the change in performance and give you feedback. I appreciate you taking your time to look at this.

josephomills:
Screenshot 2026-09-07 at 1 22 28 AM
"...Mark chapter eight and twenty two says..." only detects Mark 8:1 because the previous parse used to cater for spoken 'and' and different variations a scripture reference can be given. 


Screenshot 2026-09-07 at 1 17 30 AM
"...Ephesians cha four verse two so this the spiritual part of atmospheresians four verse two..." this truncation (chapter -> cha) and then joining of words (atmosphere Ephesians -> atmospheresians). the truncations and joins make the regex matching fail cos the regex requires full words. I was able to get these mistakes down significantly with STT rework

Another thing is the time it takes for speech to be committed (change from grey to white) which also has an impact on detection latency cos detection only happens on receiving the final utterance segment.

then also when the preacher asks for a translation, it's not possible now cos the multi-bible detection is removed. Now detection is only limited to the current tab. 

Then generally a number of detections are 'false positives'. There's no phrase or adjacency evidence required for the current system. The old matcher gated emission behind several independent floors plus bigram evidence, and had a 318-line precision suite.

also when a verse is auto projected, it only shows in the output, the suggestion doesn't show on the closed bubble and the 'show in drawer' feature is also removed so the presenter can't follow along and move to other verses if it's a multiverse reading.

josephomills:
This is a long read 😬

# What the removed pieces were doing

Notes on the mechanisms that came out during the rework, and what each was there for. Most were answers to specific failures seen in live services rather than speculative engineering, so this is mainly so you can judge which are worth having back. Several are probably not. 
👍🏽 - what I think we should have back cos it improves the feature

## Detection

**Quote matching precision.** 👍🏽 The capability is still there and is now the primary detector, so this is only about what sat around it. Emission used to be gated on several independent checks before anything surfaced: phrase and bigram evidence, minimum-match floors, and a classify step that could decline to emit at all. Those gates are what kept weak candidates from reaching the operator, and their absence is the likely cause of the unrelated references appearing in testing. A phonetic layer also sat underneath, recovering misheard biblical names that exact matching cannot.

**Verse bounds.** You've already implemented this in a different way.

**Alias verse-gating.** 👍🏽 The misheard-book-name aliases each carried a require-a-verse flag, because many of them are ordinary English: look, dude, acts, romance, judge, exit. Without the flag, normal speech triggers them.

**Misheard verse words.** The verse word accepted whisper's usual substitutions (best, this, vers, versus, worse), since "verse five" often arrives as "best five".

**Glued references.** 👍🏽 Recovered "john 316" as 3:16, using the book's real chapter count so genuinely ambiguous numbers stayed literal instead of being guessed.

**Vocabulary biasing.** Biased the decoder toward book names before a mishearing happened, rather than repairing it afterwards.

**Multi-bible search.** 👍🏽 Matched recitation against every installed translation, so a quote was found in whatever version the speaker was actually reading from.

## Performance

**Worker.** Index building and matching ran off the main thread, so neither could freeze the UI mid-service. Both now run on the main thread. Matching does yield between its widening windows, but the build does not: building the KJV index takes about 268ms of blocking main-thread work, on top of parsing the bible JSON.

**Typed arrays.** 👍🏽 The index held postings as packed integer arrays with a shared string pool, rather than Maps of Sets of strings. Measured on the real KJV (31,102 verses, 772,567 token instances, 12,576 unique words) with the current structures and tokeniser: **70.2MB**, against **4.8MB** for the packed equivalent holding the same data. Roughly fifteen times. The cost is that every verse reference is a distinct string repeated across the inverted index and both token maps, and every token instance is a separate string object.

**Memory budget.** A hard cap stopped a large library exhausting memory. Caches are keyed by bible id and never evicted, so each translation the operator switches to during a service adds its own index for the lifetime of the session — three bibles is on the order of 200MB.

## Operator control

**Output lock.** Auto-projection checked it. The lock is how an operator says nothing changes what is live.

**Manual-override pause.** 👍🏽 Detected the operator replacing a slide the AI itself had projected and paused auto-projection on that, deliberately distinguished from ordinary output use so normal operation didn't pause the feature.

**Restore previous.** Undo for an AI projection, capturing what was live beforehand.

**Follow-in-drawer.** 👍🏽 Moved the scripture view to the projected passage, so the operator kept their place and could see the surrounding verses.

**Cooldowns.** One stopped the same verse re-projecting, the other stopped a new detection cutting off a verse that had only just gone up. There is content de-duplication now, but it is skipped above 95 confidence, and a spoken reference scores 98 — so the same reference mentioned twice appears to re-project each time. The only remaining brake fires after a *manual* output change, not an automatic one.

## Plumbing

**Anchor feedback.** 👍🏽 The live passage was reported back to detection, so a bare "verse 16" resolved against what was on screen, and the matcher could follow a continuous reading.

**Book-number remapping.** 👍🏽 Bibles outside the 66-book canon number their books differently, so projection resolved the target bible's own numbering rather than assuming canon order.

**Provider error handling.**

vassbo:
Fixed "Mark chapter eight and twenty two says" not getting the verse number, one line fix.
Let me know if you have any other examples.

Multi-bible search: This is just removed temporarily.
Follow in drawer: It will always show in the drawer.
It already changes 316 to 3:16

Please look at the functionality difference, and not the code difference.

vassbo:
Made some other minor tweaks and optimizations.

Anchor feedback: This should work again.
Manual-override pause: It's already there, I just increased the timeout.

josephomills:
Thanks. I believe you're understanding the advantages of some of the old implementations now.

With the quote matching and detection, 'and' is just one of the cases. There are a lot more cases to handle for the spoken references and then some books have to be handled specially like Psalms and one chapter books (Eg Philemon, Jude). These are just a few examples:

| What's said | What's detected | Expected | |
|---|---|---|---|
| "Mark chapter 8 and 22 says" | Mark 8:22 | Mark 8:22 | ✅ |
| "give me John 316" | John 3:16 | John 3:16 | ✅ |
| "Romans 8 verses 28 through 30" | Romans 8:28-30 | Romans 8:28-30 | ✅ |
| "first John 4:7" | 1 John 4:7 | 1 John 4:7 | ✅ |
| "verse sixteen of John chapter three" | **John 3:1** | John 3:16 | ❌ |
| "John chapter 3, the 16th verse" | **John 3:1** | John 3:16 | ❌ |
| "Mark chapter eight and twenty two" | **Mark 8:1** | Mark 8:22 | ❌ |
| "the twenty-third psalm" | no match | Psalm 23 | ❌ |
| "turn to Psalms 119" | no match | Psalms 119 | ❌ |
| "Psalm 150 verse 6" | no match | Psalm 150:6 | ❌ |
| "the eighth chapter of Romans" | no match | Romans 8 | ❌ |
| "Philemon verse six" | no match | Philemon 1:6 | ❌ |


Would you want to look into it first or should port them over with the rework?

vassbo:
I still think the old system was overcomplicated for what it does. That's why I made a new system.

Thanks for the examples, I'll look into getting it to work with those.

vassbo:
All of your examples should work now to a certain extent.

josephomills:
Streaming vs batch Nemotron on 16 sermon clips that each contain a spoken reference, clean and re-recorded through a mic in a noisy room. Same model, same audio, same sherpa 1.13.7.

| Set | Driver | Word lag | WER |
|---|---|---|---|
| clean | batch | 5.8 s | 49% |
| clean | **streaming** | **0.3 s** | **47%** |
| room | batch | 6.5 s | 54% |
| room | **streaming** | **0.3 s** | **50%** |

Word lag is spoken-to-transcript, which is what detection waits on. CPU cost is the same for both. WER is high on both because the reference transcripts are whisper's own and the room set is deliberately degraded, so read the columns as relative.

Whisper is left untouched and still selectable. Once multilingual Nemotron has been tested we can look at it again.

vassbo:
Great! Looks good.

But why did you add back the HRule style prop?

josephomills:
Shouldn't have. Your cleanup removed it but AiModelManager still passed it, and I fixed the svelte-check error from the wrong side. Dropped the attribute instead.

vassbo:
No worries, I was just curious. I see that I missed the AiModelManager line earlier.

josephomills:
Cool. 

Screenshot 2026-09-07 at 12 57 03 PM

Also the floating bubble looks dull (less colourful and 'cloudy') now compared with the other existing floating buttons. I see you changed the palette and other elements. Is that the intended look?

vassbo:
Still a draft kinda. I don't want it to be too intrusive. But it can get more vibrant when it receives audio signals, you can play with it if you would like to.

josephomills:
I'll leave the bubble work for now.
Pushed a few more updates.

vassbo:
@josephomills Do you have any more plans on this specific PR, or what's the status?
I'm going to do a final cleanup, and get it merged tomorrow I guess.
Of course after this is merged you can make a new PR if there's more.

josephomills:
Okay sounds great.
For now, I may probably work on a few minor improvements. But feel free to go ahead with the merge and any beta release. Like you said, I'll just create a new PR for any more enhancements after you merge. eg. I'm waiting on my PR on sherpa to add hotwords/contextual-biasing. 

Ultimately, my plan is basically to get STT, detections and quote-matching as accurate and fast as possible. I'm coming from a background of scripture-heavy sermons, no notes before service, switching between translations fluently and we already have a skilled projection team (not to boast, lol). We've tried other dedicated AI scripture tools which haven't worked (in terms of accuracy, speed, limit on available translations & other features). So if we can get this to work well, then I can say confidently that it'll work fine for all types of congregations. And I'm personally not for building another AI scripture software, I'd rather contribute to have the feature on FreeShow cos it does much more

josephomills:
I think cross-translation quote-matching or switching translations via voice would be my next in line. 
But I'd want to know how you'd want it done, in a general sense.

Next will be handling paraphrasing.

josephomills:
I think the whisper path check has a bug.

vassbo:
> I think the whisper path check has a bug.

I know, it has not worked for a while.

josephomills:
For the cross-translation quote-matching, if any user has many translations, the app will freeze while the bibles are being indexed. That's why I had run it on a worker initially. Open to indexing in a web worker? or should I index in sequence one at a time in the background (opened translation first then favourites then the rest). Do you have any idea the average number of translations users have? people in my circles will have 15+

vassbo:
I'm pretty certain that worker you had only worked in the dev build, and would not work in a built version.
I made it load more asynchronously.

I added searching in multiple scriptures.
Currently it will search for content matches in all local scripture files, but it's not optimized to detect the differences. So you can optimize the detection confidence.

> Do you have any idea the average number of translations users have? people in my circles will have 15+

I guess most churches have just one or two translations that they commonly use.

vassbo:
Merged! Thanks for helping me with this. :)

It's at least working as a basic level, but now you can more easily experiment with the system and improve as much as you want on a seperate branch, then send a PR at any time.

Got the floating color style improved as well:
image

josephomills:
> I'm pretty certain that worker you had only worked in the dev build, and would not work in a built version.
> I made it load more asynchronously.

Great!

> I added searching in multiple scriptures.

Nice! Will check it out

> I guess most churches have just one or two translations that they commonly use.

👍🏽

> Merged! Thanks for helping me with this. :)

Exciting! 

> It's at least working as a basic level, but now you can more easily experiment with the system and improve as much as you want on a seperate branch, then send a PR at any time.

Yhup yhup. Thank you! Beta release coming soon?

gladsonsam:
Thanks for all the work on this. Can't wait to try it out!

belos117:
Installed the beta and it works fantastically on my home laptop. I'm running it live this weekend during Sunday school with a Ghanaian congregation here in Italy, off a mixer feed, and I'll report precision and recall on detected references, accented English is a case I haven't seen covered here.

One thing from setup: the only way to stop the listener mid-service is the settings tab. During a service the operator needs that in one click, a toggle and a listening indicator in the main interface would help. Is there already a status element in the drawer panel that could be promoted, or did that go in the rewrite?

vassbo:
@belos117 I was thinking it's more useful if it's on the whole service (If it's not auto presenting all the time). But I suppose it might be useful to pause it in some way.

PR #3580, opened 2026-08-06
RTMP: hardware encoding and multi-destination streaming

ip4368:
Hardware encoding
- Add per-platform encoder profiles (VideoToolbox, NVENC, QSV, AMF, VAAPI, x264). Each profile owns its full arg contribution rather than just a codec name, since -tune/-preset are x264-specific and VAAPI additionally needs pre-input device setup and an nv12/hwupload filter chain.
- Detect encoders at runtime in two stages: enumerate via `ffmpeg -encoders`, then prove each one with a real 3-frame test encode. A listed encoder still fails when the GPU or driver is absent, and static ffmpeg builds often ship NVENC/QSV without usable hardware behind them. Results are cached per binary.
- Expose an Auto / Software / specific-encoder setting, with unavailable options greyed out and their probe error shown. Fall back to x264 when a hardware encoder dies within 3s of start, so a driver fault cannot take a live service down.

Multi-destination
- rtmpData now holds a list of destinations, migrated once from the previous single url/key pair.
- One encoder emits mpegts on stdout and Node fans it out to one `-c copy -f flv` relay per destination, so N destinations cost one encode.
- Relays are isolated: each has its own buffer, and one that falls more than 2MB behind is restarted rather than dropping chunks mid-bitstream (mpegts resyncs on the next PAT/PMT). Relays reconnect with exponential backoff, and adding or removing a destination no longer restarts the encode.

Fixes
- Broadcast at the configured resolution. updateRtmpState read window.getBounds(), which getRenderBounds() halves for capture-only outputs on HiDPI displays, so RTMP had been streaming at half resolution since that DPI change landed. Use intendedBounds instead.
- Drop the `isRetina = platform === "darwin"` guess. The encoder now spawns on the first frame using its actual dimensions, so a 1x Mac display no longer produces a skewed stream, and the scale filter disappears when input and output already match.
- Stop ffmpeg children on app exit; stopAll() had no caller.
- Resolve the ffmpeg path asynchronously instead of blocking the main thread on execSync, twice per stream start.
- Push status through a registered listener rather than importing the IPC layer from RtmpStreamer, which closed an import cycle back through responsesMain and broke the existing store/files cycle at startup.
- Hold settings changed while start() is still resolving, and cancel a start that is stopped mid-flight, so neither is silently lost.

Tests: 118 total, including six driving real ffmpeg (mpegts validity, two- destination fan-out, relay add/remove, status listener, and both startup races). Not covered: non-Apple hardware encoders, which need real hardware, and the slow-destination backpressure path, which needs a throttleable server.


vassbo:
I recently fixed an RTMP audio delay issue, please look at the conflict.

ip4368:
Will check soon and update the PR, thanks @vassbo

vassbo:
I'm having issues with poor quality received by YouTube `Error - YouTube is not receiving enough video to maintain smooth streaming. As such, viewers will experience buffering.`
Also audio seems to be delayed still.

@ip4368 I'm sure it's unrelated to your changes, but would you like to take a look? :)

ip4368:
I did have the same thing as well. This is due to hardware accelerated encoding, not related to multi-destination feature.

The reason for that is the bitrate arg (-b:v) is the target bitrate and usually hardware does VBR instead of CBR, so if the content is not complex enough, it is easy to have lower bitrate than youtube recommendation but it should be relatively minor issue that shouldn't impact the quality.

I think one thing we can consider is removing -maxrate (cap), which allows some complex scene to use higher encoding and archive the target average bitrate.

Another option is that some hardware encoder allows us to set it to a CBR mode, but that also means we could be wasting bandwidth for something uncessary (to only get rid of youtube warning without material difference)

vassbo:
@ip4368 Okay, thanks. :)

PR #2790, opened 2026-02-03
feat: Enhance Quick Search with Bible/Media support and categorization

jeremiahseun:
- Update Quick Search UI to look like Mac Spotlight (wider, centered, headers).
- Implement async search logic to handle Bible search.
- Categorize results into Songs, Bible, Media, and Others.
- Add Bible search functionality (Reference and Text search).
- Add Media search functionality.
- Implement trigger actions for Bible (opens Scripture tab) and Media (plays media).

## Screenshots
Screenshot 2026-02-04 at 11 57 08
Screenshot 2026-02-04 at 11 57 30
Screenshot 2026-02-04 at 11 58 52

PR #3581, opened 2026-08-06
Show search: strict word matching, absolute match scores and near-miss fallback

### LINKED ISSUE #2107: Show search brings up irrelevant results and doesn't search show contents well
**Describe the bug**
When I search for shows, I expect on shows with the search term in the title or on slides to show up. Since lyrics will be added as shows, then I should be able to search the lyrics of a song without knowing the exact title. Currently, if I search knowing the title of the the song, the result comes with many other irrelevant search results, making the results cluttered. If I don't know the title, the show doesn't come up.

**Reproducing**
Open shows tab
search for any show using the search bar.
Analyse the search results well

Expected: only relevant shows either with the search term in the title or in a slide of the show should come up
Actual: the results are cluttered with irrelevant results. It seems to be more accurate with titles than content of the show that is not in the title.

**Screenshots (Optional)**

**Version (Optional)**
 - OS: MacOS Sequoia 15.5 
 - FreeShow Version: 1.4.9

**Additional context**

josephomills:
Fixes #2107

## Problem

Show search scores the *fraction* of query words present, using substring matching. Two consequences:

- **One matching word is enough to include a show**, so longer, more specific queries return *more* irrelevant results instead of fewer.
- **Words match inside other words**: searching `anointing here` floods the list with every show titled "There's…", because "here" is a substring of "T**here**'s". None of them contain the word "here" at all.

It also ranked "There's An Anointing Coming Over Me" above "There's An Anointing Here" for the query `anointing here`, and the match bar always showed 100% for the top result regardless of how weak it was (scores were renormalized to the best hit).

## Changes

**Matching** (`showSearchFilter`)
- A show only matches when **every** search word matches the **start of a word** in the title or content — adding words always narrows the results, and a query containing a word that matches nothing returns no results (the create-new-show hint shows instead). Partially typed words still match (`amaz grac` finds "Amazing Grace"), and the existing fuzzy title path still rescues typos (`amzing grace`).
- Quoted `"phrases"` are now word-boundary anchored too.

**Scoring** — the match bar now shows absolute confidence instead of rank:
| Match | Score |
|---|---|
| Exact title / song number / CCLI / title starts-with | 100 |
| All words in the title (adjacent words in order highest) | 75–90 |
| Strong fuzzy title match (typo) | 60–85 |
| Words split between title and content | 55–75 |
| All words in content (lyrics) only | 40–60 |

Adjacent words in order outrank scattered ones, so `anointing here` now puts "There's An Anointing Here" first at 90%.

**Search drivers** (`Shows.svelte`, `SelectShow.svelte`)
- Removed the refinement cache: it re-searched only the previous results, which were already truncated to 50/30, so a show cut by truncation could never reappear while narrowing (and the cache was unsound with the fuzzy path). Searching the full list is cheaper than before — query parsing and store reads are hoisted out of the per-show loop, the per-keystroke deep clone of the full show list is gone, and strict AND short-circuits on the first unmatched word.
- The "create new show" hint now only appears when nothing fully matches (top score < 40), so a perfect lyrics-only match no longer offers to create a duplicate on Enter.

Scripture/Bible search, media search and all other search paths are untouched.

## Tests

`search.test.ts` extended to 30 tests covering the word-boundary rule ("here" ⇏ "There's"), strict narrowing (a garbage word ⇒ no results; adding a word only ever narrows), the score bands, adjacency ranking, type-ahead prefixes, typo tolerance and quoted phrases. Full unit suite passes; svelte-check reports no issues in the touched files.

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1367,8 +1367,6 @@
         "transparent_slides": "Transparent slide preview",
         "style_template_preview": "Use style template as slide preview",
         "full_colors": "High contrast colors",
-        "search_near_miss_fallback": "Show closest matches when nothing matches the entire search",
-        "search_near_miss_fallback_tip": "Shows matching all but one word are listed with a low match when nothing matches every word. When disabled, the search is fully strict: only shows where every word matches are listed.",
         "next_item_on_last_slide": "Arrow left/right changes project item on first/last slide",
         "slide_number_keys": "Play slides with number keys",
         "auto_shortcut_first_letter": "Auto shortcut to first letter in text",
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -1367,6 +1367,8 @@
         "transparent_slides": "Transparent slide preview",
         "style_template_preview": "Use style template as slide preview",
         "full_colors": "High contrast colors",
+        "search_near_miss_fallback": "Show closest matches when nothing matches the entire search",
+        "search_near_miss_fallback_tip": "Shows matching all but one word are listed with a low match when nothing matches every word. When disabled, the search is fully strict: only shows where every word matches are listed.",
         "next_item_on_last_slide": "Arrow left/right changes project item on first/last slide",
         "slide_number_keys": "Play slides with number keys",
         "auto_shortcut_first_letter": "Auto shortcut to first letter in text",
--- a/src/frontend/components/drawer/pages/Shows.svelte
+++ b/src/frontend/components/drawer/pages/Shows.svelte
@@ -6,7 +6,7 @@
     import { activeEdit, activeFocus, activePopup, activeProfile, activeProject, activeShow, activeTagFilter, categories, drawer, focusedArea, focusMode, labelsDisabled, shows, sorted, sortedShowsList } from "../../../stores"
     import { translateText } from "../../../utils/language"
     import { getAccess } from "../../../utils/profile"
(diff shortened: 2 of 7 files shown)


vassbo:
Okay, I suppose this should help.

I don't want a search toggle `search_near_miss_fallback`, it should just work.

Did you see the existing `fastSearch` function that sits there unused? Maybe it's not needed at all.
It was added here, but I think it had some issues: https://github.com/ChurchApps/FreeShow/pull/2790

josephomills:
Alright. Then I can take the toggle off and keep strict search. 
No, I didn't see the fastSearch function. I can take a look.

josephomills:
I looked into fastSearch (searchFast.ts). The Bible part (buildBibleIndex/fastBibleSearch) is in use by Quick Search and should stay. The show part is fully dead code — fastSearch is only referenced by the commented-out line in quicksearch.ts, and the invalidateSearchIndex() calls clear an index nothing reads. It also scores with additive OR (one matching word lists a show), so re-enabling it would bring back the #2107 behaviour, and its prefix matching iterates the whole word index per term. With the strict scorer in this PR a plain scan is a few ms even on thousands of shows, so I'd say the show half isn't needed — happy to remove it here or leave that to you.

vassbo:
Great! Thanks a lot. :)

Issue #3347, opened 2026-06-06
Chords with no text overlap


Benceking24:
**Describe the bug**
When you have chords that are not linked to any text for instrumental parts of a song it is being generated overlapping.

**Reproducing**
- Create a stage layout with chords enabled, grow to fit text scaling, chords font size: 60
- Creat a show/song that has the following text:
```txt
[Intro]
[E] [Esus4] [E] [Esus4] 
```
- Trigger slide

You will see that chords are not readable because of overlapping.

**Screenshots (Optional)**
<img width="1838" height="1394" alt="Image" src="https://github.com/user-attachments/assets/651fd57c-e5b4-4e23-bab7-afca002dce32" />

**Version (Optional)**
 - OS: Window 11 Pro 25H2 (26200.8457)
 - FreeShow Version: 1.6.2-beta.1

**Additional context**
In most of the songs ChordPro syntax allows to dipslay chords without text for sections that are instrumental like song sections: intro, instrumental, breakdown, outro...

Issue #2748, opened 2026-01-27
Typing on new line moves cursor to beginning of line after first character is entered


csafarjr:
I had this issue with the previous version as well and I believe it was fixed in a beta version. But, I am having this issue again in the latest version. 
Any time I try to type in a new line it moves the cursor after the first character is typed. 

**Version (Optional)**
 - OS: Zorin OS 17 (Ubuntu derivative)
 - FreeShow Version: 1.5.6

vassbo:
I can't recreate unfortunately. You can try out the beta?

vassbo:
Okay, so I believe I got this fixed. But it was actually just moving to the start for a split second when pressing enter to create a new line or when pressing backspace to delete a line. So if you typed quickly it would place some chars at the start, but not in most cases.

Issue #2533, opened 2025-12-10
Typing in textbox moves cursor to start of line


csafarjr:
**Describe the bug**
When typing in a new line of a textbox, the cursor moves to the beginning of a line. In the same textbox, this behavior occurs on every new line as well but not after the initial typing of the line (only when the line is blank). 

**Version (Optional)**
 - OS: Linux 
 - FreeShow Version: 1.5.4

vassbo:
It's already fixed: https://github.com/ChurchApps/FreeShow/releases/tag/v1.5.5-beta.2

Issue #2490, opened 2025-12-02
Slide Textbox First Letter Cursor Switch


MyGuyConsultingYou:
**Describe the bug**
When I type my first letter in a textbox on a slide the cursor jumps to the left of the letter making my first letter the last letter when I finish typing my word.

**Reproducing**
Create a show->go to edit page->add textbox->type first letter

**Screenshots (Optional)**
https://youtu.be/u_cP8HV85w4

**Version (Optional)**
 - OS: 
 - FreeShow Version: 

**Additional context**

vassbo:
Thanks, will look into it.

MyGuyConsultingYou:
I noticed it happens whenever you go to a new line as well...the cursor moves after the first letter.

vassbo:
https://github.com/ChurchApps/FreeShow/releases/tag/v1.5.5-beta.1

PR #3582, opened 2026-08-06
Show editing fixes: backspace caret jump, style tools, group locks, history/undo

josephomills:
Fifteen focused fixes for show-editing bugs, most of them marked with WIP/TODO comments in the code. Commits are grouped in three areas.

### Text editing / caret

- **Backspace anywhere on a line moved the cursor to the start of the line.** Every Backspace forced a full HTML rebuild of the editbox (the `keyboardLineMutation` path), which destroys the native caret; `setCaret` then failed silently through one of six bare returns — worst being an unconditional bail whenever the box's last line ends in `<br>` (a trailing empty line) and the restore targets line 0. The editbox now only rebuilds when the line/segment structure actually changed (merges/splits), and `setCaret` always positions the range with clamped fallbacks instead of bailing. The guards added for #2748 / #2533 / #2490 / #1427 still apply on structural changes.
- Enter did nothing in an empty textbox, including after undoing into an empty state — `getStyleHtml` now always renders a splittable line block (resolves the WIP note at the top of EditboxLines).
- Enter/Backspace in list view stripped the following text styles — plain mode now maps styles back through identity attributes instead of positional indexes that desync after line count changes.
- Alt+Enter slide split now works in list view (resolves "TODO: get working in list view"); the selection-based `cut_in_half` shortcut is skipped while the caret is inside a quickEdit box, and only the focused editbox instance handles the split.
- Chords duplicated across styled segments (the `data-chords` partition filter's second clause was always true before the last segment) and shifted on mid-line splits (`cutLinesInTwo` reassigned whole-line chords sequentially). Both now partition by character position; relates to #3347.

### Style tools

- Pasting item style to multiple items applied styles to the wrong items — `setItemStyle` reversed `items` but not `values` (the "fix items swapping positions!" workaround was the actual mispairing bug).
- Changing a style with multiple non-text items selected replaced every item's whole style string, position included, with the last selected item's — now merges the changed key into each item's own style.
- Editing the default "Name" overlay swapped textbox contents — the overlay/template apply block now pairs each selected item with its own computed value; the find-a-non-string guard that performed the swap is gone, and lineless text items receive their style string instead of corrupting neighbors.
- Empty textboxes never received style changes (`addStyle` dropped zero-length covered nodes and restored the old style).
- Font-size on a partial selection while auto size is enabled was invisible — it now disables auto size (toast + separate undo entries) and applies the size.
- Auto size measured with forced `center`, which is half-blind to overflow (`scrollHeight` cannot see overflow above the box) — it measures with `flex-start` now; vertical alignment does not affect content size, so computed fits are unchanged for top/center.
- Small ones: slide index 0 treated as falsy in three places, `offsetTop || 5 - 5` operator precedence in the slide list autoscroll, and `$activeEdit.items` being mutated in place by sort/reverse/push.

### Locks, history/undo, stability

- **Locking a group did not lock its child slides** — children never carry `locked`. A new `isSlideLocked` helper in `helpers/show.ts` resolves the parent group and is used in Editbox, SlideEditor, quick edit (Slide), slide-style paste, and EditTools (which had its own inline version).
- **"Maximum call stack size exceeded"** (the TODO above `getTempItems`) was actually circular text-variable references: `getVariableValue` ⇄ `replaceDynamicValues` only guarded direct self-reference, so two variables referencing each other recursed on every render. A resolving-set guard returns the raw text instead of freezing. Also fixed a TypeError for stage slide offsets beyond the generated temp slides.
- History entries changing different style keys no longer coalesce into one — changing font size then color now undoes as two steps; previously the earlier change's undo data was lost (resolves "TODO: slide text edit, dont override different style keys!").
- No-change entries are skipped and no longer clear the redo stack (resolves "TODO: remove history obj if oldData is exactly the same as newdata").
- Undo support for deleting actions, variables, and interactions, and for duplicating projects. This surfaced a bug in `historyStores.ts` where `oldValue` was read off the Writable store object instead of its value, so variable-edit undo restored `undefined` — fixed. Folder duplication is documented as intentionally undo-less (recursive tree duplication has no batch history entry; per-item entries could partially undo into orphaned children — deleting the copy has undo).

### Notes for review

- Behavior notes: duplicated projects now open via the project updater's select hook (same as the other duplicate paths); an undone MIDI-receive action re-registers its listener on the next edit.
- `svelte-check` error count is identical to `dev` before and after (the 346 pre-existing errors) — no new type errors.
- Manual testing focus: backspace mid-line (especially with a trailing empty line in the box), Enter-then-type on a new line (#2748 regression case), IME composition, list-view split/merge, locked group children, chord split/merge.
- The same commits also exist as three thematic branches on my fork (`fix/editbox-caret-and-text-editing`, `fix/edit-style-tools`, `fix/history-undo-group-locks`) if smaller PRs are preferred.

### DIFF
--- a/public/lang/en.json
+++ b/public/lang/en.json
@@ -570,6 +570,7 @@
     "toast": {
         "saving": "Saving...",
         "saved": "Saved",
+        "autosize_disabled": "Auto size disabled",
         "error_media": "Could not get media.",
         "empty_styles": "No styles",
         "chapter_undefined": "Chapter {} does not exist in this book.",
--- a/src/frontend/components/edit/EditTools.svelte
+++ b/src/frontend/components/edit/EditTools.svelte
@@ -7,7 +7,7 @@
     import T from "../helpers/T.svelte"
     import { clone } from "../helpers/array"
     import { history } from "../helpers/history"
-    import { getLayoutRef } from "../helpers/show"
+    import { getLayoutRef, isSlideLocked } from "../helpers/show"
     import { _show } from "../helpers/shows"
     import { getStyles } from "../helpers/style"
     import FloatingInputs from "../input/FloatingInputs.svelte"
@@ -360,11 +360,7 @@
     let profile = getAccess("shows")
 
     $: currentShow = $showsCache[$activeShow?.id || ""]
-    $: isSlideLockedFn = () => {
-        const slideId = ref[activeSlide]?.parent?.id || ref[activeSlide]?.id
-        return !!currentShow?.slides?.[slideId]?.locked
-    }
-    $: isLocked = activeId ? false : currentShow?.locked || isSlideLockedFn() || profile.global === "read" || profile[currentShow?.category || ""] === "read"
+    $: isLocked = activeId ? false : currentShow?.locked || isSlideLocked($activeShow?.id || "", ref[activeSlide]?.id || "", currentShow) || profile.global === "read" || profile[currentShow?.category || ""] === "read"
     // $: isDefault = $activeEdit.type === "overlay" ? $overlays[activeId || ""]?.isDefault : $activeEdit.type === "template" ? $templates[activeId || ""]?.isDefault : false
     $: overflowHidden = !!(isShow || $activeEdit.type === "template" || $activeEdit.type === "overlay")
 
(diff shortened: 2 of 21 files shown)


vassbo:
Awesome! This is great, thanks. :)

A couple issues, like when copying item style and pasting the item order was reverted. I reverted it for now.
And pressing enter in an empty textbox does add the lines in the object, but they are not visible until typing anything.

PR #3591, opened 2026-08-07
Redesign/tabs

gladsonsam:
UI changes but up to you. First commit is a smaller change, second commit adds rounding:

image

image


vassbo:
Thanks. I like the idea of a new style, but not this one.

PR #3608, opened 2026-08-13
Fix chords stage display

kshairrick:
Fix trailing chord spacing and alignment, trailing chords appeared higher up than others in stage display

Multiple trailing chords were getting crammed together

Text editor was reversing the order of trailing chords when there was more than 1 trailing chord


vassbo:
Looks good, thank you! :)

vassbo:
@kshairrick Can you check this out and fix it?: https://github.com/ChurchApps/FreeShow/issues/3670

PR #3611, opened 2026-08-14
Fix keyboard shortcut issues across the app

hansli112:
Audit of the keyboard shortcut handling across the app. The shortcuts are spread over `shortcuts.ts` plus ~45 separate `<svelte:window on:keydown>` listeners, each with its own guard conditions, and most of the problems below come from those guards disagreeing with each other.

Split into eight commits, each independently reviewable.

## Fixes

**Shortcuts firing while typing.** The global handler only skipped elements with the `edit` class, but not every text field has it — Backspace in the timer popup's time input deleted the selected item, F2 started a rename. The Ctrl/Cmd branch skipped its guard entirely on macOS, so Cmd+D duplicated, Cmd+H opened History and Cmd+T toggled the panels while typing.

**`Ctrl+X` deleting the wrong thing.** `copy()`/`cut()` inferred a target from the app selection when `window.getSelection()` was empty, which is the case for a collapsed caret inside a form field. `Ctrl+X` while renaming an overlay, template, effect or project item copied *and deleted* it. `paste()` had the same hole via the macOS `Cmd+V` path.

**Non-latin keyboard layouts.** `shortcuts.ts` normalizes by physical key position, but `Drawer.svelte`, `Scripture.svelte` and `TextEditor.svelte` compared `e.key` directly. On a Cyrillic layout `Ctrl+F` opened Find & Replace through the normalized path while failing to focus the drawer search through the raw one.

**Profile access bypass.** `Ctrl+1..8` switched drawer tabs from the full list, ignoring the enabled flag, profile access and the focus-mode hidden list that `Drawer.svelte` applies when rendering. Pressing `5` opened Settings even when the profile restricts it and the top bar hides the button, and `Settings.svelte` then rendered the General tab with no check of its own. The predicates now live in `profile.ts` so the keyboard and the UI can't drift.

**Remote and controller.** `[" ", "Arrow", "Page"].includes(e.key)` never matches — `e.key` is `"ArrowRight"`, not `"Arrow"` — so only the space bar was prevented and the focused list scrolled while the slide advanced. The controller had no form-field guard at all, so arrows inside the tool `<select>` changed the drawing tool *and* advanced the slide.

**Find & replace.** The search term was compiled straight into a `RegExp`. Searching for `(` threw and was silently swallowed by the `catch`; searching for `.` replaced every character.

**Enter in popups.** Several popups could only be completed with the mouse. Popups opt in by registering their primary action rather than using a blanket handler, which would double-fire on the nine popups that already handle Enter themselves — one of which deletes shows.

**Double-firing.** `Ctrl+Arrow` on the edit page moved the edit target *and* changed the drawer sub tab. `Ctrl+L` in Preview ran before the input check and silently locked the output while typing. Single-key slide/group shortcuts fired with a popup open and with alt/shift held.

**Documentation.** The shortcuts popup was missing `Ctrl+Shift+D/N/T`, `Ctrl+Shift+Z`, `Alt+Enter`, `F8`, the presenter keys, `Ctrl+Backspace`, and Delete/Backspace — which removes the selection and was documented nowhere. `F11` is hidden on macOS, where the handler is a no-op.

## Deliberately not changed

**macOS clipboard keys.** `menuTemplate.ts` builds Cut/Copy/Paste as custom `click:` items with no `role` and no accelerator (the accelerators were removed in c08ff7f8, *"Mac shortcuts should no longer be repeated"*). That means the app's own handlers may be the only ones running for `Cmd+C/X/V` on macOS, so they stay reachable while typing there. `copy()`/`cut()` now handle the field's own selection explicitly rather than relying on native behaviour. **This path is worth a manual check on a real Mac** — I could not verify it here, as synthetic key events don't exercise Chromium's editing-command layer.

**`Ctrl+Z` in text fields.** `preventDefaults = ["z", "y"]` blocks native text undo on every platform. Pre-existing and left alone; it needs its own fix.

**Electron menu accelerators.** They look missing, but re-enabling them would re-introduce the double-firing fixed in c08ff7f8, and `mc("copy")` routes to a handler that reads `window.getSelection()`. On Windows the native menu isn't used in production at all (`index.ts` sets it to `null`).

## Verification

- `svelte-check`: 236 errors / 58 warnings / 241 hints — identical to the `dev` baseline, line for line
- `vitest`: same 7 pre-existing failures as baseline (all in `src/electron/streaming` and `utils/requests`, untouched here)
- Production frontend build and all four server bundles build
- `prettier --check` clean; `eslint` reports only pre-existing issues outside the changed lines

Empirically checked in Electron 37 / Chromium 138 that `window.getSelection().toString()` *does* return text selected inside `<input>`/`<textarea>` and is empty only for a collapsed caret — which is what pins down the actual trigger for the `Ctrl+X` bug.


vassbo:
Thanks. Please merge the in new branch so the commits are fixed.

vassbo:
?

hansli112:
Sorry for the slow reply. Redone in #3678 — new branch off current `dev`, just the eight commits and 31 files this time. The rebase was conflict-free, so the content is the same as here.

PR #3740, opened 2026-09-14
Add OnStage content provider

cabaemima:
Connect FreeShow to OnStage with OAuth (PKCE) and import services as projects and shows.

- Register an OnStage provider in ContentProviderRegistry with connect, disconnect, startup load and service sync
- Format imported lyrics and match per-service arrangements against existing shows instead of duplicating them
- Import each event into a team specific folder
- Reload a single linked song from the show context menu
- Switch the active OnStage team from Connection settings without reconnecting
- Show a sync button in the top bar while a provider is connected, and let the confirm popup offer multiple choices
- Cover formatting, song building and arrangement matching with tests


vassbo:
I feel like this adds a lot of unnesessary clutter.
Especially in `responsesMain.ts`. Also I don't think `forceReplace` is needed.
And we should not have `providerArrangement.ts`, the conversion should be done on the backend like the other providers.
Also instead of `reload_provider_show` we should use the same project refresh option that Planning Center uses.
There are many long comments.

Please simplify and improve the code.

vassbo:
Great, thanks! :)

Issue #3744, opened 2026-09-14
OpenLP forced splits get imported as plain text


efg-geislingen:
### Before submitting

- [x] I searched existing open and closed issues.
- [x] I reproduced this on the latest available version.

### Describe the bug

OpenLP has the ability to add so called "forced splits" into songs, which are represented by `[--}{--]`.
When importing songs with these splits, they are imported as plain text.

### Steps to reproduce

In OpenLP (3.1.6)
1. Create a new song
2. Add some lyrics with a forced split
3. Save the song

In FreeShow
1. Go to file > import
2. Select OpenLP/OpenLyrics
3. Select the song database to import all songs
4. Search for the song with the forced split
5. The split got imported as text

### Expected behavior

A forced split should get recognized as such and split the song text into two separate slides

### OpenLP

Song preview
<img width="415" height="191" alt="Image" src="https://github.com/user-attachments/assets/e0ee9787-f4f4-48ff-96f7-7d510c83d15d" />
Song editor
<img width="736" height="403" alt="Image" src="https://github.com/user-attachments/assets/efec85a4-5c47-4179-ba5d-bc802006ea03" />
XML in the database
<img width="1287" height="247" alt="Image" src="https://github.com/user-attachments/assets/1d2bed92-8f14-4951-b8e1-8cb5b1d95276" />  

### FreeShow
<img width="1791" height="967" alt="Image" src="https://github.com/user-attachments/assets/68c62864-e135-4045-b911-f0dc6cbdf2bf" />

### Operating System

Linux

### FreeShow version

1.6.5

### Additional context

Just stumbled upon this project recently, awesome work!

PR #3746, opened 2026-09-15
Fix OpenLP forced slide breaks during import

jokrasno:
## Summary

- Recognize OpenLP's `[--}{--]` forced-split marker during song import.
- Convert each forced section into a separate FreeShow slide while preserving existing child-slide and chord behavior.
- Keep support for the existing OpenLyrics HTML page-break marker.
- Add regression tests for inline markers, standalone marker lines, and HTML page breaks.

## Issue context

Addresses #3744. The issue was marked completed on September 15 without a linked PR, comment, or commit; the current `main` branch still only handles the HTML page-break marker. This PR supplies the missing implementation and tests.

## Testing

- `npx vitest run src/frontend/converters/openlpSlides.test.ts --config config/testing/vitest.config.ts` — 3 passing
- Production frontend build — passed (1,449 modules transformed)
- Prettier check for all three changed files — passed
- Frontend ESLint check for all three changed files — passed
- `git diff --check` — passed

For completeness, the broader Windows test run reached 111/115 passing tests. Four RTMP/ffmpeg integration tests failed outside this converter area, and `syncManager.test.ts` failed during collection in an Electron mock. Repository-wide Svelte/format checks also report existing diagnostics in unrelated files; the changed files pass their targeted checks.


vassbo:
It was already closed as completed before you opened this PR! There was a linked commit in the issue actually.

PR #3755, opened 2026-09-18
Make CCLI SongSelect footer detection independent of paste whitespace

MarkJepson:
## What this fixes

`isCCLIBlock()`/`extractCCLIMetadata()` already do the real work of pulling the CCLI Song #, writer, and copyright fields out of a SongSelect export's footer — but they only fire correctly if that footer already happens to land as its own isolated `"\n\n"`-separated section, which depends entirely on how the paste's blank lines fell.

When the writer name/footer ends up glued directly onto the last lyric line (no blank line between them — something SongSelect's own export doesn't always guarantee), the whole glued block gets treated as one section. `isCCLIBlock()` still matches it (it contains "CCLI Song"), so the **entire section — including the real lyrics — gets popped and discarded as metadata**. The last verse silently disappears from the show, and the crude "first unrecognized line is the author" fallback grabs the section's own label (e.g. `"Verse 1"`) as the author instead of the actual writer name.

## Why this approach

Rather than duplicating or changing the existing field-extraction logic, this adds one small, self-contained module (`txtSongSelectDetect.ts`) that anchors on the literal `CCLI Song #<number>` line — text no ordinary lyric paste ever contains, so it's safe to run unconditionally with no false-positive risk. From that anchor it reshapes the *input text* so the footer becomes exactly one blank-line-separated trailing section and the title line (SongSelect always puts it on its own first line, before a blank line and the first section label) is removed from the lyric body — then hands off to `isCCLIBlock()`/`extractCCLIMetadata()` completely unmodified to do the actual parsing.

Net diff to `txt.ts` itself: one import, one function call at the top of `convertText`, and one line preferring the detected title in the existing name-fallback chain. Nothing about section classification, the metadata regexes, or the UI changes.

## Verified

- Confirmed the failure mode against the current, unmodified `isCCLIBlock`/`extractCCLIMetadata` (glued footer → verse dropped, author = section label).
- Confirmed the fix live in the app against a real SongSelect export ("Way Maker") — all 6 sections present with correct lyrics, and metadata (title, author, CCLI #, copyright, year, licence) all correct.
- Confirmed no change in behaviour for exports where the footer was already isolated (the case that already worked).
- Confirmed a plain song with no CCLI footer is an untouched no-op.
- `svelte-check`, `eslint`, `prettier`: no new errors on any touched or new file, checked against a baseline diff.
- 5 new unit tests covering the glued-footer case, the already-isolated case, no-writer-line, no-CCLI-line, and a degenerate-input edge case.


vassbo:
I feel like this is a SongSelect issue more than anything, if it's not exported as a seperate block.
Also the user can just break it up if it's the wrong format. I really don't want to account for too specific scenarios.

MarkJepson:
That's fair, and I get not wanting to chase every possible paste shape.

The one thing I'd gently push back on: even if the missing blank line is arguably SongSelect's own inconsistency rather than something FreeShow caused, the current behaviour doesn't just mis-parse the footer - it silently drops the last verse's actual lyrics, with no error or warning. That felt like more than a formatting nuisance, which is why I thought it was worth a fix rather than just a "paste it correctly" note.

Totally happy to close this if you'd rather leave it as-is though - no worries either way, and thanks for taking the time to look at it.

MarkJepson:
One thing I should've mentioned in my reply above: this work was actually something you invited via email after the earlier PR — you asked me to see what I could do about detecting it more reliably, which is what prompted this branch. So I want to be clear I wasn't chasing this unprompted.

On the "SongSelect inconsistency" point — SongSelect is used by a huge number of churches worldwide, so I'd have expected its export format to be pretty consistent paste-to-paste rather than varying. If it genuinely is inconsistent, that seems like more of a reason to make the parsing resilient to it, not less, since it'll likely affect more people than just my one test case rather than being a one-off fluke.

vassbo:
I think we can probably keep "unknown" lines before the Song ID as a slide still.

I meant detecting the song parsing like Verse, Chorus more reliably. I was not talking about the SongSelect parsing.

MarkJepson:
Ahh, that goes back to my own personal project I originally emailed about converting from Python to TypeScript to include in FreeShow. Let me know if you want a bundled Python repo sent over via email.

vassbo:
Got this improved: https://github.com/ChurchApps/FreeShow/commit/fd473a0d9b8c22671981bf146b73bc25686a921f

Sorry about the confusion. If you want to take a look, I would like that you updated the current system instead of porting to account for the couple of edge-cases you found, if that's not too hard!? As if sounded like your Python scripts were much more code heavy, so porting everything would not be that easy.

PR #3756, opened 2026-09-18
Fix two section-classification edge cases

MarkJepson:
## What this fixes

Two edge cases from the earlier feedback thread, in the existing section classifier (`txtAutoSections.ts`) — not a new module, not a port, just two one-line changes.

### 1. A short one-off line could hijack the whole song's classification

`isRepetitiveRefrain()` flagged a block as a chorus refrain whenever its two most-frequent words covered ≥50% of the block's total words — a threshold that's trivial to satisfy on a short line (any 2–4 word line with one repeated or coincidentally-common word hits it). Since `findPrimaryChorusIndex()` checks this before anything positional, a short one-off line (an intro, an ad-lib) could get anchored as the "primary chorus," and every other section in the song would then cascade to "bridge."

**Fix:** require at least 6 words before the ratio check runs. A short one-off line can no longer trip it, while genuine repetitive refrains — which run several words long by nature — are unaffected.

### 2. Mixed labelled/unlabelled songs mislabelled trailing verses as bridges

Once any section matched an explicit header (e.g. `[Chorus]`), `passedInstrumentalBreak` was set permanently, which disabled the verse-footprint check for the rest of the song. In a song mixing explicit labels with plain unlabelled verses, any unlabelled verse appearing after the first label fell through to "bridge" instead of "verse," even when it structurally matched verse 1.

**Fix:** verse-footprint matching no longer depends on that flag — it already only fires for blocks that structurally match verse 1, which is signal enough on its own.

## Verified

- Both original failure cases from the earlier report: a short-line intro no longer hijacks `primaryChorusIndex` or cascades other sections to bridge; a mixed labelled/unlabelled song now correctly classifies trailing plain verses as verse.
- Two cases that already worked correctly (a normal Pre-Chorus/Chorus/Bridge song, and a chorus repeat with minor ad-libs) are unchanged.
- `svelte-check`: 598 errors/61 warnings/3742 hints — identical before and after, checked against a stashed baseline.
- `eslint`: clean.
- Net diff: 2 lines changed, nothing else touched.


MarkJepson:
I hope this is what you meant :)

vassbo:
Great, this is good. :)

